PR #368, opened 2015-07-28
Similar to 366, DialCallStatus should use failed instead of FAILED

peteroyle:
Again, not tested, but DialActionTest:211 looks for all money like it will still work:

assertTrue(data.getFirst("DialCallStatus").equalsIgnoreCase("failed"))


gvagenas:
Thanks Peter.

PR #378, opened 2015-08-07
Combine both implementations of getNetwork()

peteroyle:
I'm finding that, out of the two differently implemented versions of getNetwork(), one works for me (on CentOS) and one doesn't. This PR replaces the one that doesn't work with the one that does.


peteroyle:
Suspected there might be a reason there are two different implementations of this and think I see why: The existing one works on Ubuntu only, and the proposed one works on CentOS only. I'll see if I possess enough bash skillz to come up with something that works on both...

peteroyle:
Completed, and created #385

PR #388, opened 2015-08-18
sync master

pchlupacek:


pchlupacek:
This has been discussed with jean.deruelle@telestax.com this is initial request for comments on bringing akka / scala up to current mainstream release version. 

Please not this is WIP, and I would like to get comments now primarily on 
 - `Pre23Props` workaround hence the way how the actors are created in current master is not anymore available in akka 2.3 onwards
- Disregards please all specialised `Creator` code in actors, this was initial attempt how to resolve issue, and I am gonna to remove it. 
- https://github.com/Mobicents/RestComm/pull/388/files#diff-ddffdfc91fa631668885f73ea7fac1f1L813 and similar where the explicit target of actor is provided, is this really `self`?

Also can you please suggest the correct way to run all tests so they pass on master as green? I cannot really get to that point, and want to be sure I am not doing something wrong. 

After I will finalise this work, I am gonna to close this PR and open new one for the final review. 

Thanks for understanding 

Pavel.

Issue #421, opened 2015-08-30
in getPublicIP() failover to observebox.com when ipecho.net not available


peteroyle:
ipecho.net has had some availability issues over the last couple of days causing the RestComm instances in our test environment to not start up. I am currently testing a patch which will try to get the public IP via observebox.com if ipecho.net is unavailable at the time.

peteroyle:
The attached PR is currently working for us even though ipecho.net is down at the moment.

PR #422, opened 2015-08-30
#421 If ipecho.net is not responding, try observebox.com

peteroyle:
As per #421, if ipecho.net is does not respond in 5 attempts, try observebox.com for 5 attempts.

Something else that's also different to the current implementation, only 10 attempts will be made overall and if they fail then the script will exit with a failure message, rather than hanging indefinitely.


gvagenas:
HI @peteroyle , the http://www.observebox.com/ip returns a 404 Not Found for me. Can you please update the script with the proper url?
Thanks

peteroyle:
Hi @gvagenas. Looks like http://observebox.com/ip is no longer in service so I replaced it with http://ip-addr.es

PR #424, opened 2015-09-01
Implement tests for desired fallover URL behaviour across 3 scenarios

peteroyle:
I have created 3 new tests in the RestComm test suite to capture the way I would like RestComm to behave under these 3 conditions:
- If Voice URL returns a 404, try the Fallback Voice URL straight away
- If the Voice URL contains a hostname which doesn't resolve either, try the Fallback URL (whether straight away, or after a timeout of maximum 5 seconds)
- If the Voice URL times out, try the Fallback Voice URL within 5 seconds

See also the discussion in RestComm Google Group called 'Making more use of failover URL'.


peteroyle:
Created ticket #468 to track this.

Issue #468, opened 2015-09-27
Try Fallback Voice URL when Voice URL returns 404


peteroyle:
Hi,

As per google group discussion at: https://groups.google.com/forum/#!topic/restcomm/KxIl8BCmkOQ

When RestComm receives a 404 from the Voice URL response it simply forwards that 404 to the caller. I would like to see RestComm instead failed over to the Fallback Voice URL in that situation. This would make the Fallback Voice URL useful in the following scenarios:

- HTTP proxy used to serve the RCML is down
- HTTP proxy is up but all application servers behind it are down
- HTTP proxy is up, application servers are also up, but the RCML application is not deployed
- IncomingPhoneNumber is misconfigured with a problematic Voice URL, but the Fallback Voice URL works

All of the above scenarios return a 404 which would currently result in RestComm ending the call, rather than attempting the Fallback Voice URL.

I have created 3 new failing tests in the RestComm test suite to capture the way I would like RestComm to behave under these 3 conditions:
- If Voice URL returns a 404, try the Fallback Voice URL straight away
- If the Voice URL contains a hostname which doesn't resolve, try the Fallback URL (whether straight away, or after a timeout of maximum 5 seconds)
- If the Voice URL times out, try the Fallback Voice URL within 5 seconds
These tests can be found in the pull request here: #424 

cc @leftyb

gvagenas:
relates to #479

peteroyle:
Hi Folks. Is there any chance this will make it into 7.6.0 before the code freeze on April 8th please? I have just had a good look at https://github.com/leftyb/RestComm/commit/07ecc47a8e1eb825b462e82fee0dd0d7a52a9394 and it looks right to me. Do we know yet whether this commit makes the tests I provided pass?

gvagenas:
@peteroyle the fallback url is already supported and this one here is a duplicate. Can you please test the latest Restcomm and report here?

peteroyle:
Thanks @gvagenas I'll try it out soon. Can you please tell me where to find the other ticket which this duplicates?

gvagenas:
PR https://github.com/RestComm/Restcomm-Connect/pull/528 closes this issue

peteroyle:
Hi @gvagenas, you are right. I tested the fallback URL in the latest build and it's working fine. Thanks!

PR #498, opened 2015-10-06
Try Fallback Voice URL when Voice URL returns 404.  Issue #468

### LINKED ISSUE #496: No transition could be found from a(n) stopping state to a(n) failed state.
https://gist.github.com/leftyb/08a68a83c04680b52c46

When calling 1238, from a BOB registered SipPhone, canceling the call from an Alice registered SipPhone, provokes this error.

leftyb:
Hi @peteroyle these are the modifications for the issue, that you opened. Please test them and comment for anything if necessary, 

*Added a timeout = 5 secs (hard coded) for the "DefaultHttpClient" for downloading RCML, in case that the hostname can not be resolved.

*As well fixes  #496


peteroyle:
Thanks @leftyb I'll give it a try as soon as possible and let you know how I go.

Issue #87, opened 2014-10-21
Multi-tenancy support


ivelin:
Currently there is no API to create a new top level account.
I had a conversation with George and it turned out that all accounts share the same assets.
For example account A can see all DIDs of account B and the other way around.
The following behavior should be implemented:
Peer level accounts should not see any of the other account's assets.
2. Parent accounts should see all of a children's account assets.
3. A sub-account/child should not see the parent assets

ivelin:
Another issues came up that is related to multi-tenancy. Since all users login with the same ID and share assets (numbers and apps), over time the UX becomes a mess. Noone knows which apps and numbers are stale and can be removed. 

Each user should have their own login and assets to manage. Only admins should be able to see and manage apps and numbers across users, but it should be clear which assets belong to whom.

ivelin:
Another issues came up that is related to multi-tenancy. Since all users login with the same ID and share assets (numbers and apps), over time the UX becomes a mess. Noone knows which apps and numbers are stale and can be removed. 

Each user should have their own login and assets to manage. Only admins should be able to see and manage apps and numbers across users, but it should be clear which assets belong to whom.

deruelle:
It seems the isolation is working in a better way for the newest version where a sub account can't see the parent DID but the applications are still visible to everyone. It seems this depends on https://github.com/Mobicents/RestComm/issues/4 to be completed in both RestComm and RVD

ivelin:
Thanks for the update @deruelle . Its important to close this issue for the SaaS launch.

otsakir:
Yes, the application notion should be supported by core restcomm. Currently we are treating applications as URLs. It's quite limiting.

@ivelin , regarding multi-tenancy support the issues you mention are partly dealt with on the SSO branch and will be solved for good with the release.

ghjansen:
By the last commit for this issue we have the following access scenario:

**Accounts API**: an account can manage its own info and also info from its immediate sub accounts only.
**Applications API**: an account can manage its own applications only.
**IncomingPhoneNumbers API**: an account can manage its numbers and the numbers from its immediate sub accounts only.

Having in mind that we need to focus on Roles in soon, as mentioned on issues #272 and #440, maybe the implementation of this issue could be reviewed to make sure it will not conflict with, or cause unexpected behaviours after Roles became fully functional. I can do this ;)

ghjansen:
[...]

PR #499, opened 2015-10-06
Issue4 applications rest api

ghjansen:
This branch includes modifications related to Issues #4 and #87 .


gvagenas:
@ghjansen I sent email with comments, questions and suggestions for this PR, email subject "PR #499: Application Rest API (Issue 4) and Multitenancy (Issue87)"

Thanks

otsakir:
@gvagenas , regarding:

"The field org.mobicents.servlet.restcomm.entities.Application.rcmlUrl why we need this new field since we have already the voiceUrl ( and voiceFallbackUrl, smsUrl, smsFallbackUrl) ?"

I've discussed the issue with @ghjansen . The conclusion was that the relation of an Application entity and an RVD project should be 1-1 so that each RVD project/RAS package etc. can be treated individually. For example, it would be nice if we could use the Application.friendlyName as the application name. But if the Application contains three different Urls (voice, sms and ussd) that means it is bound to three different RVD projects. It doesn't make sense. The same applies for all application attributes that currently exist (or will be added) to the Application entity. In order to have any meaning they should refer to a single RVD project. 

That leaves us with two options:
- remove all *urls and keep only rcmlUrl together with a new "kind" attribute that is one of "voice|sms|ussd". I guess that's the path Gui is on.
- keep voiceFallbackUrl, smsUrl, smsFallbackUrl .. and all other *Urls in the Application and change the RVD project so that it can host functionality for voice/sms/ussd etc in a single project.

The easiest i think is the first option and also seems more flexible.

gvagenas:
@otsakir @ghjansen Its clear now, we should go with the first option:
*  remove all *urls and keep only rcmlUrl together with a new "kind" attribute that is one of "voice|sms|ussd". I guess that's the path Gui is on.

and maybe rename the field "url" to "appUrl" or "rcmlUrl" as you suggest so its clear that this field points to the url we should use to download the app rcml.

Also, is there a plan for deeper integration between RVD and Restcomm on the Application API, or its already there? For example when user creates RVD application, automatically create Application entity at Restcomm side - or at least provide user with the option to create Restcomm Application entity.

And last, Admin UI needs any work for the new Application API?

George

ghjansen:
Thanks @gvagenas and @otsakir , here follows additional info:

*Why is needed the method org.mobicents.servlet.restcomm.http.AbstractEndpoint...*
 * This method was created to ensure that the credential used to access the API will be able to manage his own information and from its sub accounts, and not able to manage information from parent accounts or accounts at the same level. This was applied to the same API's used by UI.

*line 135 What is the case where (!sidPrincipal.equals(accountSid)) are not the same?*
 * If you manually create a request informing the login credentials of a parent account, and the accountSid of a sub account in the URL, this will be the mentioned scenario. To handle this case, we need to ensure that the information requested is inside the hierarchy. So secureLevelControl() is basically preventing invalid combinations between the sidPrincipal (login credentials), accountSid (url informed) and referenceAccountSid (the account that owns a requested DID, Client, ... - if one of these were informed) based on the tree this components need to form when backtracking from its owner sid.

*line 142, what if a parent account wants to list Incoming Phone Number that he and his child accounts own?...*
* As I mentioned above, this checks are only preventing the invalid combinations. So if a parent account (login credential - principal sid) try to list his child information (url - accountSid) it will works. This is valid to a list of DIDs and also if the request points to a specific DID from the sub account, but this DID needs to be owned by the informed sub account, that needs to be owned by the parent account informed as login credentials (phew!)

*The field org.mobicents.servlet.restcomm.entities.Application.kind should better be an Enum...*
 * I agree. I almost made it using Enum, but was in doubt if real necessary. I can do it anyway, no problem.

*Application entity (org.mobicents.servlet.restcomm.entities.Application) should have support for Ussd applications*
 * This could eliminate 'rcmlUrl' and 'kind' attributes, but conflict with @otsakir previous mentions.

*Currently IncomingPhoneNumbers have VoiceURL field for the RCML. Adding the Applications*
  * I agree. In previous talks with @otsakir we discussed this possibility, but is good to have in mind that Twilio allows both possibilities: inform a plain URL, or an existent Application. When a new call arrives, Restcomm is already checking which one is present on IncomingPhoneNumber entity to create the call.

*... and maybe rename the field "url" to "appUrl" or "rcmlUrl" ...*
 * I think the current name is 'rcmlUrl' already. Can you inform if you find something different?

*... is there a plan for deeper integration between RVD and Restcomm on the Application API ...*
 * I've implemented replication to Applications API when RVD Projects are created, renamed, imported or deleted. This is already in this PR.

*And last, Admin UI needs any work for the new Application API?*
 * Admin UI is not currently using Applications API, but we can do it by improving the way we link DIDs with Applications.

gvagenas:
@ghjansen see my comments bellow:

1. Probably I miss something for the method: org.mobicents.servlet.restcomm.http.AbstractEndpoint.secureLevelControl(AccountsDao, String, String). 
Using the 
```Line 134: String sidPrincipal = String.valueOf(SecurityUtils.getSubject().getPrincipal());``` you will get the current logged in username.
```Line 135: if (!sidPrincipal.equals(accountSid)) ``` we check if the accountSid is equal to logged in user accountSid, but this will be always true. I don't see how we "...ensure that the credential used to access the API will be able to manage his own information and from its sub accounts, and not able to manage information from parent accounts or accounts at the same level...". Can you please elaborate on that?
What if we use REST API from the command line, such as curl to access REST API, in that case the getSubject().getPrincipal() will be the same as the credentials used for Curl request.

2. Please proceed to : "...The field org.mobicents.servlet.restcomm.entities.Application.kind should better be an Enum.."

3. For the Application entity please proceed per Orestis suggestion : "...remove all *urls and keep only rcmlUrl together with a new "kind" attribute that is one of "voice|sms|ussd". I guess that's the path Gui is on...". Also 

4. Agree on your comments for "..Currently IncomingPhoneNumbers have VoiceURL field for the RCML. Adding the Applications.." so keep both Application SID and url to be compliant with Twillio. Make sure also that we support fallback Applications or urls for Voice, Sms and Ussd,

George

ghjansen:
Thanks @gvagenas

1. The operation sidPrincipal.equals(accountSid) will be always true as you said when the API is accessed by the AdminUI only, once that the JavaScript layer assembles the requests this way. But using Curl as you also mentioned, things may become different.
Lets assume the existance of a Primary Account that have a Subaccount A. So if the primary try to access the list of DID's of this sub account using Curl, we'll have sidPrincipal ≠ accountSid, as I simulate below:

Primary Account sid = ACae6e420f425248d6a26948c17a9e2acf
Subaccount A sid = AC1ec5d14c34a421c7697e1ce5d4eac782

curl -X GET http://ACae6e420f425248d6a26948c17a9e2acf:[authToken]@localhost:8080/restcomm/2012-04-24/Accounts/AC1ec5d14c34a421c7697e1ce5d4eac782/IncomingPhoneNumbers.json

When this request reach the server, we'll have the following:

sidPrincipal == 'ACae6e420f425248d6a26948c17a9e2acf'
accountSid == 'AC1ec5d14c34a421c7697e1ce5d4eac782'

This is the way that they will be different. Even then, the request will be allowed because the relationship between sidPrincipal with accountSid in this scenario is parent-children, so its ok, the parent account will be able to see his child DIDs list.

Now if the Primary Account try to use Curl to get the DIDs List from the Default Administrator (primary's parent) or from a Secondary Account (same level as primary), we'll also have sidPrincipal ≠ accountSid, but an AuthorizationException will be throwed since none of Default Administrator or Secondary Account have Primary account as parent, as verified here:

https://github.com/Mobicents/RestComm/blob/Issue4_ApplicationsRestApi/restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/AbstractEndpoint.java#L137

So, this is the way we should be able to let accounts "be able to manage his own information and from its sub accounts, and not able to manage information from parent accounts or accounts at the same level." Please, let me know if you find anithing strange on this, or if you wish that I evolve even more with examples or even test it in a specific scenario.

2. 3. 4. Will do! After end with the modifications, should I push to the same branch and create another PR?

Thanks!

ghjansen:
@gvagenas a754fdca5c1691c0ecf17511331be4aa2d2a42bf has the last modifications you suggested on this PR. Summary changes:
- Application.kind as enum (Application.Kind);
- Applications.\*url and Applications.\*method removed from entity structure;
- Interpreters CallManager, SmsService and UssdCallManager adapted to the new structure (Using rcmlUrl from the application **if** the DID has a applicationSid, otherwise using the url from DID. Fallback and callback urls will be always from the DID.)
- Application entity related tests were adapted.

My suggestion to near future is to improve the the way we link DIDs with Applications by the AdminUI, to start setting the IncomingPhoneNumber.\*applicationSid instead of keep the url only, and have a strongest data model. But we probably need to study some impacts on that also.

Please, let me know if there is something else on this. Thanks!

gvagenas:
@ghjansen 

1. You haven't merge with master and you didn't resolved the conflicts. From inside your branch you need to run "git merge master" and then "git mergetool" and take care the changes needed
2. DB IncomingPhoneNumbers are still using URLs. We should turn some of the entries to Applications and tie the nunber to the application and keep some numbers with url.

George

gvagenas:
@ghjansen Also we will need to change the Admin UI integration with RVD. Currently AdminUI will ask RVD to get the list of available applications and will create incoming phone number with url.

Since part of your work is for RVD to use Application API to create and delete apps, we need Admin UI to lookup for apps from Restcomm.
That means, that at Admin UI whenever someone creates a new number, we should query Restcomm with account SID to get the list of available applications for this account and use them to create the new number.

PR #634, opened 2015-11-10
Test whether <Dial callerId> is honoured when from-address-to-proxied-calls is true

peteroyle:
Attempt at defining what the caller ID should look like when callerId attribute is specified on a Dial tag, even though \<from-address-to-proxied-calls\> is set to true.


peteroyle:
Created issue #635

gvagenas:
@peteroyle thanks for the PR and the test you provided. 
I create a patch for this, check issue #775 .
Now iff CallerId is provided in the Dial verb Restcomm will use it no matter what the "from-address-to-proxied-calls" option is set

PR #837, opened 2016-02-02
784 Issue - Allow Restcomm client authentication with the user ID and password of a Restcomm account user

thinhly197:


deruelle:
@thinhly197 one comment, I see that the new code doesn't provide a non regression test case ? Can you add one ?
Also did you solve the secure storage of the password in the clients Database ?

thinhly197:
@deruelle Yep, I can add a regression test. I intended to do it, but I can't find where is the place to implement it. Could you please advise me?
About secure storage of the client password, I did it. But I think the another Restcomm client application also need to encrypt the password before compare with one from database (e.g. Olympus)

deruelle:
@thinhly197 in https://github.com/RestComm/RestComm-Core/tree/master/restcomm/restcomm.testsuite. 
Yeah that's what I meant, can you check if as part of the issue you can also encrypt the passwords for that table.

thinhly197:
@deruelle I think it is not the part of this issue. That why I proposed we can use the default password is client's login name. The user can change it later and it doesn't have an impact to the account's password

deruelle:
So you actually use the same password for the account and for the client when it's created, right ? 
Can you create a separate issue for the secure storage of the client password ?

thinhly197:
I still use MD5 to encrypt the client password for this commit. I will update it and commit new one with regression test.
Yes, I will create new issue for the secure storage.

Issue #829, opened 2016-01-29
The sample RVD apps don't show up as 'linked' to the respective numbers in Admin UI incoming numbers screen


atsakiridis:
Seems that the RVD apps are linked in reality, but don't show up as linked in the admin UI. As a result, when you make a call to a number (i.e. 1241) that appears in admin ui as not connected, you still get the gather RVD app. The other issue is that if you try to update the RVD app from the admin UI (for example change 1241 and 'link' it to RVD say demo) it still points to the old. 

As a result: 
- some numbers that are supposed to be linked to RVD show up unlinked and 
- an app is connected when calling the number from olympus and that app cannot be updated in admin ui

Here are the logs:

https://drive.google.com/file/d/0BwfD74mG87iMak5XdXF0VzFLdVk/view?usp=sharing

otsakir:
@atsakiridis , after some investigation i noticed the following:

DIDs to demo RVD applications are created using the following query in restcomm.xml:
```
INSERT INTO "restcomm_incoming_phone_numbers" VALUES('PN46678e5b01d44973bf184f6527bc33f7','2014-02-17 22:37:08.709000','2014-02-17 22:37:08.709000','This is an IVR app that maps user input to specific action','ACae6e420f425248d6a26948c17a
9e2acf','+1240','2012-04-24',FALSE,NULL,'POST',NULL,'POST',NULL,'POST','AP73926e7113fa4d95981aa96b76eca854',NULL,'POST',NULL,'POST',NULL,'/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/IncomingPhoneNumbers/PN46678e5b01d4
4973bf184f6527bc33f7',NULL,NULL,NULL,NULL,TRUE,'0.0')
```

New DIDs, when mapped to an RVD applicationcontain no AP* sid foreign key. For example:
```
INSERT INTO "restcomm_incoming_phone_numbers" VALUES('PN8a4ea0898dfb4b85bd40752bae2d4e59','2016-02-01 14:18:51.259000','2016-02-01 14:18:51.259000','4455','ACae6e420f425248d6a26948c17a9e2acf','455','2012-04-24',FALSE,'/restcomm-rvd/servic
[...]

PR #844, opened 2016-02-02
Reverting the use of Applications on DIDs to URLs

ghjansen:
Related to issue #829.


gvagenas:
SQL Update script:  https://github.com/RestComm/RestComm-Core/blob/master/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/changelog/mariadb/update_script_Feb2_2016.sql

PR #893, opened 2016-02-22
#9 added prompt-uri and cache-uri for XMS

jhermanski:
Added prompt-uri and cache-uri to config-restcomm.sh for use with Dialogic XMS audio prompts. There is a corresponding change in the RestComm-Docker project (restcomm_service.sh) for issue https://github.com/RestComm/Restcomm-Docker/issues/9.


hrosa:
@jhermanski I was reviewing commit 443c1f16fc2a2df4d58da171b2783034139b21f3 and I see you're in a good path, but unfortunately the addresses and port cannot be hardcoded like that.

For example, if the Media Server is deployed in a different server than RestComm, then the ${bind_address} will not allow the Media Server to access the audio files hosted on RestComm server.

My advise is: if the STATIC_ADDRESS variable is defined, then RestComm should use it. Otherwise, use relative URL (RestComm is able to decode relative URLs, so no need to bother here).

Another issue is the port. You're hardcoding 8080 port which is fine for HTTP. **But what if RestComm is configured to use HTTPS?** I advise reading the DISABLE_HTTP variable. If false you use HTTP:// with 8080 port; if true you use HTTPS:// with 8443 port

One last thing: we also need to configure the recording path and URL.

Let me know if this is clear.

Thanks

leftyb:
@hrosa @jhermanski  
The necessary modifications for patching prompt-uri and cache-uri is done and merged.
Please see this [PR](https://github.com/RestComm/Restcomm-Connect/pull/1002)

PR #959, opened 2016-03-24
issue600

atika2k:
Restcomm redirects to the number in format "+1234". When registering a number in AdminUI, RestComm considers "+1234" and "1234" as same numbers.

### DIFF
--- a/restcomm/restcomm.telephony/src/main/java/org/mobicents/servlet/restcomm/telephony/CallManager.java
+++ b/restcomm/restcomm.telephony/src/main/java/org/mobicents/servlet/restcomm/telephony/CallManager.java
@@ -305,7 +305,8 @@ public final class CallManager extends UntypedActor {
         // TODO Enforce some kind of security check for requests coming from outside SIP UAs such as ITSPs that are not
         // registered
 
-        final String toUser = CallControlHelper.getUserSipId(request, useTo);
+        String toUser = CallControlHelper.getUserSipId(request, useTo);
+        
         final String ruri = ((SipURI) request.getRequestURI()).getHost();
         final String toHost = ((SipURI) request.getTo().getURI()).getHost();
         final String toHostIpAddress = InetAddress.getByName(toHost).getHostAddress();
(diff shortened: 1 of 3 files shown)


gvagenas:
@atika2k Thanks for your work and your help to the project.
I reviewed your work and unfortunately it is not ready to merge with master.

The request here is for Restcomm to treat numbers 1234 and +1234 the same way. 
So if we have registered number 1234, and someone calls +1234, Restcomm should match the number and execute the RCML that is attached to this.

Your patch adds the (+) at a number if missing, and does that only if the client != null.

The proper patch should check if the number contains the (+) and check the DB for a number with (+) if nothing found, check again without the (+).
1. New call arrives to Restcomm for number +123456789
2. Restcomm will check if DB contain number +123456789. If number found proceed as usual
3. If number at step 2 is not found, remove + and search for number 123456789 and proceed accordingly.

The above checks should be done at: https://github.com/RestComm/Restcomm-Connect/blob/65c7f4ca8909b44c045c6ece0f587e67e954b4c9/restcomm/restcomm.telephony/src/main/java/org/mobicents/servlet/restcomm/telephony/CallManager.java#L495

Thanks
George

atika2k:
Hi George,

The changes i've made in controllers/numbers-incoming.js (Admin UI) allow you to register a number only with "+". If you try to register a number ex. 123456789, "+" is added automatically. It means the number that will be stored in the DB is with "+" (+123456789). 

When a client dials a number without + , then the CallManager adds "+" and then checks the DB for that number.
If i understand you correctly a client should check the DB twice before initiating a call. But if it is registered with "+", then there is no need to check twice.

Please correct me if i am wrong.

Regards,

Dejan

deruelle:
@atika2k I'll let @gvagenas  comment but please note developers can register numbers via the REST API as well, not only through Admin UI. http://docs.telestax.com/restcomm-api-incomingphonenumbers/

deruelle:
Thanks @atika2k for the updates. @gvagenas will review shortly.

atika2k:
Hello @gvagenas,
Any updates on the review?

gvagenas:
Hi @atika2k.

I reviewed your last work and you are on the right path for this patch, I have 2 more comments though:

1. You should remove the AdminUI (numbers-incoming.js) patch you added there. The user should be able and free to register any number they want, either with (+) or not. Restcomm should later try to match the incoming phone number to a number in the DB - so your patch at CallManger is ok

2. The test you provided doesn't properly test the patch. The proper test should have the following properties:

  - Have a number registered in the database WITH the (+) sign, for example +123456789, and that uses application with just Play verb, see https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.testsuite/src/test/resources/hello-play.xml. Next a sip client calls to the number WITHOUT the (+), that is calls to 123456789. Restcomm should answer the call (100 - 180 - 200 OK) and few secs later Restcomm will disconnect the call
  - Have another number registered in the database WITHOUT the (+) sign, for example 987654321, and that uses application with just Play verb, see https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.testsuite/src/test/resources/hello-play.xml.. Next a sip client calls to the number WITH the (+), that means calls to +987654321. Restcomm again should answer the call (100 - 180 - 200 OK) and few secs later Restcomm will disconnect the call

Let me know if you need any help with the test.

Thanks
George

PR #969, opened 2016-03-29
Issue#859 patch duplicate statuscallback action url duplicate status

croufay:
patch duplicate statuscallback action url duplicate status


deruelle:
@croufay please add a non regression test case. It's missing from the PR.

Issue #637, opened 2015-11-10
RestComm Geolocation Rest API Implementation


deruelle:
Implement Design specified in #636

deruelle:
This should include tests in the restcomm.testsuite that Emulate the GMLC MLP (since it is HTTP Based you can use wiremock as we do in DID Provisioning Unit tests) so we don't need to do any work yet on GMLC until the API is implemented and tested.

FerUy:
Closed with #1013

gvagenas:
Open until review and merge of PR #1013

ghjansen:
Hi @FerUy and @gvagenas , here follow some considerations from my review for this PR. 
@FerUy, please consider that my knowledge about Geolocation particularities is really limited, so feel free to contest and elaborate on arguments about the content below.

To improve the visualization of all endpoints in Geolocation API, i created the following table considering all possible formations in the way i understood as intended.
@FerUy please check if you see some invalid combination and we adapt this post as needed.

Base path | Type | Sufix | Method | Response  
----------|------|-------|--------|---------
/Accounts/{accountSid}/Geolocation | n/a | n/a | GET | XML List
/Accounts/{accountSid}/Geolocation.json | n/a | n/a | GET | JSON List
/Accounts/{accountSid}/Geolocation | /Immediate | n/a | GET | XML List
/Accounts/{accountSid}/Geolocation | /Immediate | n/a | POST | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate.json | n/a | GET | JSON List
/Accounts/{accountSid}/Geolocation | /Immediate.json | n/a | POST | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | GET | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | POST | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | PUT | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | DELETE | HTTP 204
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | GET | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | POST | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | PUT | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | DELETE | HTTP 204
[...]

PR #1013, opened 2016-04-21
Geolocation API

FerUy:
This is a pull request for issue #637.
Find documentation of Restcomm Geolocation API definition (with examples) and RCML here:
https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api/geolocation-api.adoc
https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/rcml/geolocation-rcml.adoc


deruelle:
I don't see the docs in asciidoc format at https://github.com/RestComm/Restcomm-Connect/tree/GeoLocationAPI/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api or https://github.com/RestComm/Restcomm-Connect/tree/GeoLocationAPI/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/rcml ?

FerUy:
@deruelle I'm taking care of the docs right now :)

FerUy:
@deruelle I'm taking care of that right now :)

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/DaoUtils.java
```diff
@@ -95,6 +97,14 @@ public static Integer readInteger(final Object object) {
         }
     }
 
+    public static BigInteger readBigInteger(final Object object) {
+        if (object != null) {
+            return (BigInteger) object;
```
jaimecasero: not sure what is added value here,apart from isolating the casting... this may be designed differently, i will keep an eye here

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
```diff
@@ -0,0 +1,213 @@
...
+        super();
+        this.sessions = sessions;
+    }
+
+    @Override
+    public void addGeolocation(Geolocation gl) {
+        final SqlSession session = sessions.openSession();
+        try {
+            session.insert(namespace + "addGeolocation", toMap(gl));
+            session.commit();
```
jaimecasero: are you expecting compound dao operations scenarios? commiting here assumes add is always using its own transaction. It may work now, but it maybe you need to add, and update in the same Tx, so this approach could be problematic. I would manage sessions from an interceptor, and save session in localthread, so the Tx demarcation is more flexible.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
```diff
@@ -0,0 +1,213 @@
...
+        } finally {
+            session.close();
+        }
+    }
+
+    @Override
+    public List<Geolocation> getGeolocations(Sid accountSid) {
+        final SqlSession session = sessions.openSession();
+        try {
+            final List<Map<String, Object>> results = session.selectList(namespace + "getGeolocations", accountSid.toString());
```
jaimecasero: avoid unccesary string concatenation. Could you create a constant since there is no dynamic part here??

jaimecasero: do we need to call toString() here. Also toString is better to be kept as diagnosis purposes like logging, maybe better to create a method that returns whatever you need..

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
```diff
@@ -0,0 +1,213 @@
...
+            }
+            return geolocations;
+        } finally {
+            session.close();
+        }
+    }
+
+    @Override
+    public void removeGeolocation(Sid sid) {
+        removeGeolocations(namespace + "removeGeolocation", sid);
```
jaimecasero: can we create constant here?

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    @Override
+    public void removeGeolocations(final Sid accountSid) {
+        removeGeolocations(namespace + "removeGeolocations", accountSid);
```
jaimecasero: constant..

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
```diff
@@ -0,0 +1,213 @@
...
+        final SqlSession session = sessions.openSession();
+        try {
+            session.update(namespace + "updateGeolocation", toMap(gl));
+            session.commit();
+        } finally {
+            session.close();
+        }
+    }
+
+    private Map<String, Object> toMap(Geolocation gl) {
```
jaimecasero: maybe you can create a class to isolate this two method loc<->map

also map keys should be constants and reused

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
```diff
@@ -0,0 +1,676 @@
...
+import java.net.URI;
+
+import org.joda.time.DateTime;
+import org.mobicents.servlet.restcomm.annotations.concurrency.NotThreadSafe;
+
+/**
+ * @author fernando.mendioroz@telestax.com (Fernando Mendioroz)
+ *
+ */
+public class Geolocation {
```
jaimecasero: is this class final?

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
```diff
@@ -0,0 +1,676 @@
...
+            throw new IllegalArgumentException(glt + " is not a valid application GeolocationType.");
+        }
+
+        @Override
+        public String toString() {
+            return glt;
+        }
+    };
+
+    public Geolocation setSid(Sid sid) {
```
jaimecasero: I know final classes are generally good in Java, but it maybe too much for a DTO class...
What is the design criteria to set this class as final?

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
```diff
@@ -0,0 +1,676 @@
...
+        }
+
+        public void setFormattedAddress(String formattedAddress) {
+            this.formattedAddress = formattedAddress;
+        }
+
+        public void setLocationTimestamp(DateTime locationTimestamp) {
+            try {
+                this.locationTimestamp = locationTimestamp;
+            } catch (Exception exception) {
```
jaimecasero: what is the purpose of this catch?

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
```diff
@@ -0,0 +1,676 @@
...
+        public void setEventGeofenceLatitude(String eventGeofenceLat) {
+            this.eventGeofenceLatitude = eventGeofenceLat;
+        }
+
+        public void setEventGeofenceLongitude(String eventGeofenceLong) {
+            this.eventGeofenceLongitude = eventGeofenceLong;
+        }
+
+        public void setRadius(Long radius) {
+            if (geolocationType.toString().equalsIgnoreCase("Notification")) {
```
jaimecasero: better to use enum comparison if availabe rather than string comparison...
Also im not sure this business logic should be placed here, it seems a bit obscure

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/GeolocationList.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+
+import org.mobicents.servlet.restcomm.annotations.concurrency.NotThreadSafe;
+
+/**
+ * @author fernando.mendioroz@telestax.com (Fernando Mendioroz)
+ *
+ */
+@NotThreadSafe
+public final class GeolocationList {
```
jaimecasero: what is this wrapping class adding?

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Sid.java
```diff
@@ -34,7 +34,7 @@
     private final String id;
 
     public enum Type {
-        ACCOUNT, APPLICATION, ANNOUNCEMENT, CALL, CLIENT, CONFERENCE, GATEWAY, INVALID, NOTIFICATION, PHONE_NUMBER, RECORDING, REGISTRATION, SHORT_CODE, SMS_MESSAGE, TRANSCRIPTION, INSTANCE
+        ACCOUNT, APPLICATION, ANNOUNCEMENT, CALL, CLIENT, CONFERENCE, GATEWAY, INVALID, NOTIFICATION, PHONE_NUMBER, RECORDING, REGISTRATION, SHORT_CODE, SMS_MESSAGE, TRANSCRIPTION, INSTANCE, GEOLOCATION
```
jaimecasero: can you associate the Codes for each enum value, so you dont need to use the literals everywhere?

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/AbstractEndpoint.java
```diff
@@ -99,6 +101,46 @@ protected URI getUrl(final String name, final MultivaluedMap<String, String> dat
         return uri;
     }
 
+    protected Integer getInteger(final String name, final MultivaluedMap<String, String> data) {
+        Integer integer = null;
+        if (data.containsKey(name)) {
+            integer = new Integer(data.getFirst(name));
```
jaimecasero: use Integer.valueOf rather than creating a new Integer

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
```diff
@@ -0,0 +1,752 @@
...
+        dao = storage.getGeolocationDao();
+        accountsDao = storage.getAccountsDao();
+        configuration = (Configuration) context.getAttribute(Configuration.class.getName());
+        configuration = configuration.subset("runtime-settings");
+        super.init(configuration);
+        final GeolocationConverter converter = new GeolocationConverter(configuration);
+        final GsonBuilder builder = new GsonBuilder();
+        builder.registerTypeAdapter(Geolocation.class, converter);
+        builder.setPrettyPrinting();
+        gson = builder.create();
```
jaimecasero: is this created object thread safe???

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
```diff
@@ -0,0 +1,752 @@
...
+        xstream = new XStream();
```
jaimecasero: this is thread-safe and may be shared as stated in doc ;)

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
```diff
@@ -0,0 +1,752 @@
...
+        xstream.alias("RestcommResponse", RestCommResponse.class);
+        xstream.registerConverter(converter);
+        xstream.registerConverter(new ClientListConverter(configuration));
+        xstream.registerConverter(new RestCommResponseConverter(configuration));
+    }
+
+    public Response putGeolocation(final String accountSid, final MultivaluedMap<String, String> data,
+            GeolocationType geolocationType, final MediaType responseType) {
+        try {
+            secure(accountsDao.getAccount(accountSid), "RestComm:Create:Geolocation");
```
jaimecasero: this security concern may be placed in another class, have you considered https://jersey.java.net/documentation/latest/filters-and-interceptors.html

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
```diff
@@ -0,0 +1,752 @@
...
+            secureLevelControl(accountsDao, accountSid, null);
+        } catch (final AuthorizationException exception) {
+            return status(UNAUTHORIZED).build();
+        }
+        try {
+            if (!geolocationType.toString().equalsIgnoreCase("Immediate")
```
jaimecasero: can be replaced by enum comparison??

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
```diff
@@ -0,0 +1,752 @@
...
+                    && !geolocationType.toString().equalsIgnoreCase("Notification")) {
+                validateGeolocationType(data);
+            }
+        } catch (final NullPointerException exception) {
```
jaimecasero: can we replaced NPE catch by if logic??

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
```diff
@@ -0,0 +1,752 @@
...
+                return null;
+            }
+        }
+    }
+
+    private Geolocation createFrom(final Sid accountSid, final MultivaluedMap<String, String> data,
+            Geolocation.GeolocationType glType) {
+        // *** Validation of not null and specific formatted parameters *** //
+        // *** Source can not be null ***/
+        try {
```
jaimecasero: I think all this validation code should be isolated in another place, this method is getting big.... Also some parts of the validation may be designed to reuse code

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/converter/GeolocationConverter.java
```diff
@@ -0,0 +1,530 @@
...
+    protected void writeSource(final String source, final JsonObject object) {
+        if (source != null) {
+            object.addProperty("source", source);
+        } else {
+            object.add("source", JsonNull.INSTANCE);
+        }
+    }
+
+    protected void writeDeviceIdentifier(final String deviceIdentifier, final HierarchicalStreamWriter writer) {
+        if (deviceIdentifier != null) {
```
jaimecasero: are thses null checks neccesary? is the invoker aware in case null??

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/converter/GeolocationConverter.java
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Override
+    public boolean canConvert(final Class klass) {
+        return Geolocation.class.equals(klass);
+    }
+
+    @Override
+    public void marshal(final Object object, final HierarchicalStreamWriter writer, final MarshallingContext context) {
```
jaimecasero: there shoul be a better way to get a JSON from that object,is not the jsx-rs providing this?

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * @author fernando.mendioroz@telestax.com (Fernando Mendioroz)
+ *
+ */
+public class RestcommGeolocationsTool {
+
+    private static RestcommGeolocationsTool instance;
+    private static String geolocationsUrl;
+
+    public static RestcommGeolocationsTool getInstance() {
```
jaimecasero: is this singleton actually required? It seems just the geolocationsUrl att is sharing state.. But is rewritten on every private method that uses it. Consider converting that att into local var, and remove the singleton...

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
```diff
@@ -0,0 +1,244 @@
...
+        if (instance == null) {
+            instance = new RestcommGeolocationsTool();
+        }
+        return instance;
+    }
+
+    private String getGeolocationsUrl(String deploymentUrl, String accountSid, Boolean xml) {
+        if (deploymentUrl.endsWith("/")) {
```
jaimecasero: this removeSlahAtTheEnd logic is duplicated, could we create a private method with this logic...?

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
```diff
@@ -0,0 +1,244 @@
...
+            deploymentUrl = deploymentUrl.substring(0, deploymentUrl.length() - 1);
+        }
+        geolocationsUrl = deploymentUrl + "/2012-04-24/Accounts/" + accountSid + "/Geolocation";
```
jaimecasero: reconsider the string concatenation strategy to build URL. JAX-rs builder may provide better ways to do it, if not simply use StringBuilders. The API version is repeated multiple times as literal, constant???

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/DaoUtils.java
FerUy: Noted @jaimecasero .... Please notice all methods throughout this class follow the same design (so, I stuck with it)

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
FerUy: Noted @jaimecasero. Again, if we agree this approach could bring trouble, then we should study all MybatisXDao classes, as this one follows the same design of all of them (adapted to its own needs of course).

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
FerUy: @jaimecasero same comment as previous. I could, but for now I would rather stick to the same design pattern used throughout these classes.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
FerUy: Yes... but as told in previous comment, for now I would rather sticking to the same design pattern used throughout these classes.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
FerUy: Identical comment as previous for this one too.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/dao/mybatis/MybatisGeolocationDao.java
FerUy: Same feedback as previous in this regard @jaimecasero

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
FerUy: Spot on @jaimecasero. Indeed this class must be final. Will fix in next commit.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
FerUy: Sorry @jaimecasero, could you expand about being too much for a DTO class?
 For the design criteria, I followed the one used for entity classes here.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
FerUy: If a non DateTime value for this attribute is received in a POST or PUT (or from a third party like a GMLC), an exception is thrown.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
FerUy: Agreed on both comments @jaimecasero
For the first one it's an easy fix... For the second one, I didn't find another way so far. I'm accepting an incoming "radius" value in PUT request in an Immediate type of Geolocation, but as it does not apply at all, it's ignored rather than being rejected and loosing potentially proper location information.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/GeolocationList.java
FerUy: The capacity of getting a list of geolocations? Sorry, I'm not sure what you are asking for here.
Please keep in mind that for any entity in this package, a class like this one is added (same for Gateway and GatewayList, Client and ClientList, etc.)

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Sid.java
FerUy: Sorry @jaimecasero, I didn't get this one (?).
Anyway, I just added the Geolocation code in this class (GL).

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/AbstractEndpoint.java
FerUy: It's definitely a better choice @jaimecasero (again, should be followed for the rest of the methods that stick with this criteria).

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
FerUy: Sorry @jaimecasero , which doc are you referring to?

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
FerUy: This is reuse and adaption of software too, so your suggestion's scope is for the rest of the team ;)

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
FerUy: Yes, already done and working OK after new Junit tests (will push with other enhancements tomorrow). All of those have been replaced. Good catch, thanks @jaimecasero

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
FerUy: Agreed @jaimecasero, actually I've been thinking how to revamp it without breaking anything before the PR. Surely it will be done very soon.

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/converter/GeolocationConverter.java
FerUy: Again, this is an adaptation of reused software @jaimecasero. All XxxxConverter in this package classes follow the same design, so I stuck with it (and initially, learnt from those).

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
FerUy: Agreed, same comment as previous one though.

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
FerUy: Although it's again reuse and adapted software, in this case, being duplicated for Notification and Immediate types, it gets so awfully replicated I indeed implemented it as you suggest. Done, tested and working fine (will be included in next push). Thanks @jaimecasero

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
FerUy: Again, this is an adaptation of reused software. However, I agree with you in the constant for API version and account and replaced it instead of this: "/2012-04-24/Accounts/" in that string. Thanks @jaimecasero

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
jaimecasero: but input param is already DateTime, there is no conversion, right?

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Sid.java
jaimecasero: there is a "generate" method later that could have benefited a lot to use the Enum string value, so no switch would be neccesary....

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/GeolocationList.java
jaimecasero: ok, i just try to prevent unccesary wrapping classes around collections, specially when using generic, i guess this is following established design patter in Restcomm ;)

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
jaimecasero: Well DTO is a design pattern usually used in conjunction with DAO pattern. I just assumed you were using it, but apparently your Entities are more VOs than DTOs. In that case, having final classes is mandatory. Funny thing is equals/hashCode is not implemented properly for all of them so they are actual VOs. Apparently there is a mix of DTOs and VOs there, anyway i can see this is the established Design again ;)

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
jaimecasero: i just checked original doc for that class

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
FerUy: Right, there's no conversion. However, the input could be rubbish, then the response informs about a rejected POST or failed PUT.

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/GeolocationList.java
FerUy: Exactly :)

restcomm/restcomm.dao/src/main/java/org/mobicents/servlet/restcomm/entities/Geolocation.java
FerUy: Yeap ;) ... thanks Jaime :+1:

restcomm/restcomm.http/src/main/java/org/mobicents/servlet/restcomm/http/GeolocationEndpoint.java
FerUy: :+1:

restcomm/restcomm.testsuite/src/test/java/org/mobicents/servlet/restcomm/http/RestcommGeolocationsTool.java
jaimecasero: In this case, I honestly think you should take the effort (which should be minimum) and prevent the Singleton, its bad for performance and testability. In addition, im not sure that static geolocationsUrl is properly proptected against concurrent access (have you tried using this tool concurrently in your JUnit test???). If this is not exactly under geoloc Scope, file an issue against it...

FerUy: As commented in Slack channel with @deruelle and @gvagenas , an issue will be created as it is widespread in all tools classes, even though this are only for testing purposes. Thanks @jaimecasero

ghjansen:
Hi @FerUy and @gvagenas , here follow some considerations from my review for this PR. 
@FerUy, please consider that my knowledge about Geolocation particularities is really limited, so feel free to contest and elaborate on arguments about the content below.

To improve the visualization of all endpoints in Geolocation API, i created the following table considering all possible formations in the way i understood as intended.
@FerUy please check if you see some invalid combination and we adapt this post as needed.

Base path | Type | Sufix | Method | Response  
----------|------|-------|--------|---------
/Accounts/{accountSid}/Geolocation | n/a | n/a | GET | XML List
/Accounts/{accountSid}/Geolocation.json | n/a | n/a | GET | JSON List
/Accounts/{accountSid}/Geolocation | /Immediate | n/a | GET | XML List
/Accounts/{accountSid}/Geolocation | /Immediate | n/a | POST | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate.json | n/a | GET | JSON List
/Accounts/{accountSid}/Geolocation | /Immediate.json | n/a | POST | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | GET | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | POST | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | PUT | XML Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid} | DELETE | HTTP 204
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | GET | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | POST | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | PUT | JSON Element
/Accounts/{accountSid}/Geolocation | /Immediate | /{sid}.json | DELETE | HTTP 204
/Accounts/{accountSid}/Geolocation | /Notification | n/a | GET | XML List
/Accounts/{accountSid}/Geolocation | /Notification | n/a | POST | XML Element
/Accounts/{accountSid}/Geolocation | /Notification.json | n/a | GET | JSON List
/Accounts/{accountSid}/Geolocation | /Notification.json | n/a | POST | JSON Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid} | GET | XML Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid} | POST | XML Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid} | PUT | XML Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid} | DELETE | HTTP 204
/Accounts/{accountSid}/Geolocation | /Notification | /{sid}.json | GET | JSON Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid}.json | POST | JSON Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid}.json | PUT | JSON Element
/Accounts/{accountSid}/Geolocation | /Notification | /{sid}.json | DELETE | HTTP 204

Assuming the table above as the intended scenario, the follow adjustments are required in the current implementation:

- **1** - Remove `@DELETE` support for `/Immediate`.
- **2** - Switch method used in `@GET` for `/Immediate` to return a list of elements, or remove `@GET` support for this endpoint.
- **3** - Remove `@PUT` support for `/Immediate`.
- **4** - Remove `@DELETE` support for `/Immediate.json`.
- **5** - Switch method used in `@GET` for `/Immediate.json` to return a list of elements, or remove `@GET` support for this endpoint.
- **6** - Remove `@DELETE` support for `/Notification`.
- **7** - Switch method used in `@GET` for `/Notification` to return a list of elements, or remove `@GET` support for this endpoint.
- **8** - Remove `@PUT` support for `/Notification`.
- **9** - Remove `@DELETE` support for `/Notification.json`.
- **10** - Switch method used in `@GET` for `/Notification.json` to return a list of elements, or remove `@GET` support for this endpoint.
- **11** - Check how multi-tenancy must be applied to `Geolocation` records (currently, only the owner has access to its own records).
- **12** - Refactor method `GeolocationEndpoint.validateGeolocationType(final MultivaluedMap<String,String>)`, replacing the current parameter by `Geolocation.GeolocationType` placing the validation of this parameter inside the method (currently, the method is only throwing an exception instead of actualy perform the validation and then throw an exception - if needed).
- **13** - Refactor the current validation of parameters when processing requests to create a new `Geolocation` entity, avoiding error responses in the same format of the successfull ones. Currently, the method `GeolocationEndpoint.validateCreateParams(final Sid, final MultivaluedMap<String,String>, Geolocation.GeolocationType)` returns a `Geolocations` object independent of the validation result, so in case of fail, the response will contain a `Geolocation` entity in JSON or XML format and one of its attributes will carry the error message. THe problem in this approach is that there is no correspondent error in the protocol layer (a HTTP error code) and the response contains attributes that should be generated with the entity just in case of success, like `sid`, `date_created`, `date_updated` and `uri`, so if the one who requested this operation may think that the `sid` in the response represents a database record, when actually it was only the attribute of a temporary `Geolocation` object, created in memory only. The following suggestions are based in the standards used accross all endpoints, and may help to adjust the current approach (the order of the items must be considered):
	- **13.1** turn the method `GeolocationEndpoint.validateCreateParams(final Sid, final MultivaluedMap<String, String>, Geolocation.GeolocationType)` into `void`;
	- **13.2** move the validation of the `geolocationType` from the method `GeolocationEndpoint.validateGeolocationType(final MultivaluedMap<String, String>)` to method `GeolocationEndpoint.validateCreateParams(final Sid, final MultivaluedMap<String, String>, Geolocation.GeolocationType)` (check item **12** first);
	- **13.3** remove the method `GeolocationEndpoint.validateGeolocationType(final MultivaluedMap<String, String>)`;
	- **13.4** rename the method `GeolocationEndpoint.validateCreateParams(final Sid, final MultivaluedMap<String, String>, Geolocation.GeolocationType)` to `validate`;
	- **13.5** remove `logger.info()` calls inside `validate` method to avoid log inflation and preserve server's filesystem;
	- **13.6** replace each **`return`** in `validate` method by a **`throw new`** followed by a convenient exception capable to transport the information and error message required to include in the response (in case of fail);
	- **13.7** remove unnecessary use of **`try`** and **`catch`** clauses in `validate` method;
	- **13.8** remove method `Geolocation.rejectGeolocationRequest`;
	- **13.9** remove method `Geolocation.buildIncorrectGeolocationRequest`;
	- **13.10** remove method `Geolocation.failedGeolocationRequest`;
	- **13.11** remove method `Geolocation.buildIncorrectGeolocationUpdateRequest`;
	- **13.12** remove method `Geolocation.incorrectGeolocationUpdateRequest`;
	- **13.13** change the logic in `Geolocation.putGeolocation` to call the method `createFrom` after call the method `validate`, so `createFrom` will be reached only if there is no validation errors;
	- **13.14** remove the existent `createFrom` method and rename method `buildGeolocation` to `createFrom`;
	- **13.15** in 'putGeolocation' method, replace the logic based on `geolocation.getResponseStatus` by a `return` which must use a HTTP error code based on `validate` result, ensuring that the validation block is placed before the creation of the `Geolocation` entity, like in the `putX` method present in other endpoints;
	- **13.16** review and apply the previous suggestions also for method `GeolocationEndpoint.updateGeolocation` and the structures under it.
- **14** - The method `MybatisGeolocationDao.removeGeolocations` seems not used, may consider to remove this method.

@FerUy some things are complex to explain, so please feel free to ping me if something is not clear.
@gvagenas your considerations are also very welcome here when you manage some free cycles.

FerUy:
Commit 5204ff8 included:
1 - Removed @DELETE support for /Immediate.
2 - Removed @GET support for this endpoint.
3 - Removed @PUT support for /Immediate.
4 - Removed @DELETE support for /Immediate.json.
5 - Removed @GET support for this endpoint.
6 - Removed @DELETE support for /Notification.
7 - Removed @GET support for this endpoint.
8 - Removed @PUT support for /Notification.
9 - Removed @DELETE support for /Notification.json.
10 - Remove @GET support for this endpoint.
11 - No multi-tenancy check to Geolocation records apply currently (only the owner has access to its records).
12 - Validation of Geolocation.GeolocationType now is performed inside method: "private void validate(final MultivaluedMap<String, String> data, Geolocation.GeolocationType glType)
            throws RuntimeException".
13 - Validation for creating a Geolocation (POST parameters) now are performed inside method: "private void validate(final MultivaluedMap<String, String> data, Geolocation.GeolocationType glType)
            throws RuntimeException".
14 - Removed logger.info() calls inside validate method.
15 - Removed unneeded us of try and catch clauses inside validate method.
16 - As now exceptions are thrown for not API compliant POST requests, instead of generating a speciall response with a Sid but no record, removed methods Geolocation.rejectGeolocationRequest, Geolocation.buildIncorrectGeolocationRequest.
17 - For geolocation requests that apply generating a record but have validation issues (e.g. wrong geographic coordinates format), the following method builds the response (with no Geolocation Data and response status / cause information): "private Geolocation buildFailedGeolocation(final Sid accountSid, final MultivaluedMap<String, String> data, Geolocation.GeolocationType glType)". 
18 - In Geolocation.putGeolocation, createFrom method is called after validate method, hence it will be reached only if there no validation errors exist.
19 - In putGeolocation method, replaced the logic based on geolocation.getResponseStatus by a return using an HTTP error code based on validate result (via throwing NullPointerException, IllegalArgumentException or UnsupportedOperationException), then avoiding creation of Geolocation entity for malformed requests.
20 - After a design decision, a malformed update will not delete any record, thus update logic for failed response status requests is maintained (although improved from coding point of view).
21 - For conflict resolving, now GeolocationEndpoint extends SecuredEndpoint instead of AbstractEndpoint.
22 - MybatisGeolocationDao.removeGeolocations is kept for further use.

As stated, still pending a fix for Junit test suite methods affected by changes mentioned in 19 (all assertions are working except for POST requests generating Exceptions, which are demanding a refactor for  methods testCreateAndGetImmediateGeolocation() and testCreateAndGetNotificationGeolocation() in GeolocationEndpointTest class).

FerUy:
Closing this pull request to avoid confusion with the project's new branch.

Issue #1014, opened 2016-04-23
InstanceID Host is Null and a new InstanceID is created on every startup on Mysql


croufay:
mysql: [Warning] Using a password on the command line interface can be insecure.
+------------------------------------+---------------------+---------------------+------+
| instance_id                        | date_created        | date_updated        | host |
+------------------------------------+---------------------+---------------------+------+
| ID27c34632c727486cb8e195bfeb467144 | 2016-04-22 21:03:52 | 2016-04-22 21:03:52 | NULL |
| ID7af4cd554d48459593cc0d58b418c245 | 2016-04-22 20:39:59 | 2016-04-22 20:39:59 | NULL |
| ID89eb9eb501534ced91f826e701634243 | 2016-04-22 12:47:15 | 2016-04-22 12:47:15 | NULL |
| IDb702aa9f7ebb400ea908574ddc4cb18b | 2016-04-22 11:08:30 | 2016-04-22 11:08:30 | NULL |
| IDcde4818002694fda9fde1fd6fc856476 | 2016-04-22 20:00:16 | 2016-04-22 20:00:16 | NULL |
| IDedb2a7ca0b864f90a79cd242738cf07b | 2016-04-22 19:37:39 | 2016-04-22 19:37:39 | NULL |
| IDfbc0e15c7ca340de837689cae4563d60 | 2016-04-22 11:04:10 | 2016-04-22 11:04:10 | NULL |

The server has multiple IP address

PR #1015, opened 2016-04-23
Issue#1014 InstanceID null host IP fix and missing variable in mariadb sql

croufay:
InstanceID null host IP fix and missing variable in mariadb sql


gvagenas:
@croufay Thanks for the patch, very good way to use the UDP Sip Connector. I used that part of your patch and closed issue #1014 .

The PR you have here contains patches for other issues also and cannot be merged like this. Please create a PR for each issue.

Also make sure you update with master, the PR here cannot be merged because of merge conflicts.

Issue #1036, opened 2016-05-04
RVD is not handling Project/Application name conflicts properly under MySQL


ghjansen:
Create a new RVD Project present a unstable behaviour if the project name was already used before by another Application of the same user Account, what results in the removal of the original Application entity on API side leaving an orphan project folder inside RVD Workspace.

The timeline bellow shows a practical example of how this issue happens:

1. Account user@company.com creates project **test** using RVD;
2. RVD request RestComm to create a new Application using Applications REST API;
3. RestComm creates a Application entity named **test** and return the applicationSid  `APb70c33bf0b6748f09eaec97030af36f3` to RVD;
4. RVD creates a new project folder inside the workspace named `APb70c33bf0b6748f09eaec97030af36f3`;
5. Account user@company.com tries to create a new project named **Test** (or **test**) using RVD;
6. RVD request RestComm to create a new Application using Applications REST API;
7. RestComm search for Applications named **Test** (or **test**) and locates the Application `APb70c33bf0b6748f09eaec97030af36f3`, because MySQL is case insensitive as default, so once this Application is from the same owner who is requesting the creation of a new Application, the response will contains all the information from Application `APb70c33bf0b6748f09eaec97030af36f3` instead of a HTTP 409 (Conflict);
8. RVD receives the response from RestComm, extract the applicationSid `APb70c33bf0b6748f09eaec97030af36f3` from it and fails to create the project folder, since there is already a project folder with the same name inside the workspace;
9. RVD handles the fail of creating the project folder inside the workspace by rolling back the creation of the Application entity on RestComm, sending a HTTP DELETE for `APb70c33bf0b6748f09eaec97030af36f3`. This removes the Application entity created on step 3, but the project folder named `APb70c33bf0b6748f09eaec97030af36f3` still remains inside the workspace as orphan.

Section 1 to 4 represents the normal workflow. Section 5 to 9 represents the critical scenario.

ghjansen:
Fixed by #1037.

Issue #1028, opened 2016-04-30
RVD Workspace Migration uses wrong Application entity to sync


ghjansen:
As noticed by the upgrade executed on cloud by @leftyb , applications with similar names can cause problems during the synchronisation with RVD Projects. This happens using MySQL database, when the migration algorithm select a application by its friendly_name, and another application with the same name is returned, given MySQL DB is case insensitive by default.

The following shows part of the migration log, where **'Test'** application was successfully migrated generating its respective applicationSid (used to rename the project folder inside the workspace), and a different application named **'test'** which was wrongly assumed as related to the application entity created to project **'Test'**:

    --------------------------------------
    Project 'Test' synchronized with Application 'AP89eca7f092694b2589951e53bc2e6085'
    Project 'Test' renamed to 'AP89eca7f092694b2589951e53bc2e6085'
    No IncomingPhoneNumbers found to update with Application 'AP89eca7f092694b2589951e53bc2e6085'. Skipped
    No Clients found to update with Application 'AP89eca7f092694b2589951e53bc2e6085'. Skipped
    --------------------------------------
    Project 'test' synchronized with Application 'AP89eca7f092694b2589951e53bc2e6085'
    Error while migrating project 'test' [ERROR-CODE:5] Error while renaming the project 'test' to 'AP89eca7f092694b2589951e53bc2e6085'Destination '/mnt/nfs/home/rvd/workspace/AP89eca7f092694b2589951e53bc2e6085' already exists
    --------------------------------------

The **BINARY** operator can be used in MyBatis selector to workaround the problem and turn the query into case sensitive: http://dev.mysql.com/doc/refman/5.7/en/charset-binary-op.html

ghjansen:
Fixed by #1037.

PR #1037, opened 2016-05-04
Issue1028_FixRVDUpgrade

ghjansen:
This PR contains fixes for two different problems.

- **Problem 1**: RVD Workspace Migration uses wrong Application entity to synchronise RVD projects, like described in detail by Issue #1028;

- **Problem 2**: Create a new RVD Project present a unstable behaviour if the project name was already used before by another Application of the same user Account, like described in detail by Issue #1036.

-----------------

- Fix for **Problem 1**: affects only the `applications.xml` file and may does not require a new build of the whole project (since a text file), this may handy to manually apply the fix inside the environments without the need of a upgrade;

- Fix for **Problem 2**: affects file `ApplicationsEndpoint.java` and depends on `applications.xml` to work properly, so may its better to trust this into a new build/compilation of RestComm.

This way, considering the PR #1027 which is able to fix one RVD Workspace Migration issue using a SQL script, it may be possible to consider also the Fix for **Problem 1** described here as one patch to fix the urgent issues without a new build of RestComm applying it manually (once they are text files), and later provide a new build with everything in place.

### DIFF
--- a/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/applications.xml
+++ b/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/applications.xml
@@ -16,7 +16,7 @@
   </select>
 
   <select id="getApplicationByFriendlyName" parameterType="string" resultType="hashmap">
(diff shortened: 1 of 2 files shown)


gsaslis:
@ghjansen doing some housekeeping here... is this still a valid PR ?

PR #1096, opened 2016-05-21
Issue748_Organization

ghjansen:


gsaslis:
@ghjansen doing some _more_ housekeeping here... is this still a valid PR ?

PR #1126, opened 2016-05-30
fix #1124

### LINKED ISSUE #1124: cache-uri is wrong for external media server options
Hi 

i have found a bug: `cache-uri` doesn't contains restcomm `port`
in case if external mediaserver is specified 

https://github.com/RestComm/Restcomm-Connect/blob/d663828440f51a5687ea78aad0caa0cc94daf5be/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh#L430

https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh#L436

thanks,

hamsterksu:


leftyb:
Hi @hamsterksu,

Sorry for the delayed response.
The modification looks good but I can see a small redundancy.
According to that: 
https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/start-restcomm.sh#L161

you dont need to do that:
https://github.com/RestComm/Restcomm-Connect/pull/1126/commits/45772e67804571a53465749334e44751c6abf28b#diff-142b73a0e53f971caf699c5e20911335R434


You can directly use:

``REMOTE_ADDRESS="${SCHEME}://${STATIC_ADDRESS}:${PORT}"``

what do you think?

Regards.

hamsterksu:
@leftyb to long config file :) so i have added one more check to be sure that everything is ok.
so your point `STATIC_ADDRESS` can not be empty. am i right? if yes i will remove `if` statement

Another point for discussion:

I don't think that it's good idea to use `"$MS_ADDRESS" != "$BIND_ADDRESS"` because in this case we can not start different instances of `restcomm` and `mediaserver` on the same machine for dev/test purpose.

leftyb:
Hi @hamsterksu,

Well the point is that STATIC_ADDRESS is not empty. If it is not set manually on restcomm.conf it takes the value of the BIND_ADDRESS, if it set use it. So yes the ``IF`` statement is not necessary. Use STATIC_ADDRESS.

On the last part, did you actually test it? to run the external RMS and RC in the same host using this configuration (using ``$MS_ADDRESS" != "$BIND_ADDRESS``)? It should run normally as well with no problem. Due to that it is like using ResComm-connect binary. So it should work.
I need to test as well. 


Regards

hamsterksu:
hi @leftyb i got your point.

you are right if `$MS_ADDRESS" != "$BIND_ADDRESS` it will work with default settings(listen localhost).

i will update my PR

hamsterksu:
@leftyb updated.

anyway i still think that `$MS_ADDRESS" != "$BIND_ADDRESS` can be removed too. in case it's the same host we will provide correct settings too. (yes it will be as default) 
as for me script will be clearer: if `MS_ADDRESS` exists that do this.

leftyb:
Thank you @hamsterksu,
Done.

leftyb:
Ok @hamsterksu I understand your point will review again and test a bit. 

Thank you.

Issue #1143, opened 2016-06-02
Newly created accounts should appear in the creator account menu


otsakir:
When a new sub-account is created it does not immediately appear in the creator account menu. One has to refresh to see it.

muhammadbilal19:
@otsakir I fix this issue and created pull request. Please review and let me know if you see any problem in the fix.

PR #1144, opened 2016-06-03
Fix issue of newly created account not shown without refresh ref: #1143

muhammadbilal19:
Please review changes for newly created sub account not shown in the menu without refreshing the page. #1143


deruelle:
Thanks @muhammadbilal19 ! @gvagenas please review

otsakir:
@muhammadbilal19 your PR has been merged. 

I've $emitted an event instead of directly calling getAccountsList() through the $scope to achieve looser coupling between the two controllers.

Thanks for the fix!

muhammadbilal19:
@otsakir thats good suggestion and thanks for changing for updating this change in code 👍

PR #1201, opened 2016-06-18
SIP diversion header available in RVD

antonmry:
It allows to have the value of SIP DIversion Header in a new RVD variable named $core_diversion. Also, it publish the variable $core_ForwardedFrom with the number so it may be used (only if the variable is empty).

We have tested with Voxbone, returning the following values:

- $core_diversion: \u003csip:YYXXXXXXXXX@voxbone.com\u003e;counter\u003d1;reason\u003dUSER-BUSY"
- $core_ForwardedFrom: YYXXXXXXXXX

A sipp scenario I've used to test it: https://gist.github.com/antonmry/cbb9383e4199fe294cbcb01d6ba1e573

### DIFF
--- a/restcomm/restcomm.interpreter/src/main/java/org/mobicents/servlet/restcomm/interpreter/VoiceInterpreter.java
+++ b/restcomm/restcomm.interpreter/src/main/java/org/mobicents/servlet/restcomm/interpreter/VoiceInterpreter.java
@@ -1137,7 +1137,7 @@ public final class VoiceInterpreter extends BaseVoiceInterpreter {
                     final String sipCallId = lastResponse.getCallId();
                     parameters.add(new BasicNameValuePair("DialSipCallId", sipCallId));
                     parameters.add(new BasicNameValuePair("DialSipResponseCode", "" + statusCode));
-                    processCustomHeaders(lastResponse, "DialSipHeader_", parameters);
+                    processCustomAndDiversionHeaders(lastResponse, "DialSipHeader_", parameters);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


deruelle:
Thanks @antonmry ! @gvagenas will review

otsakir:
@antonmry, i'm reviewing your patch from RVD side. I found no reference of  *$core_diversion* variable inside RVD. Am i missing sth?
*$core_ForwardedFrom* is fine.

restcomm/restcomm.interpreter/src/main/java/org/mobicents/servlet/restcomm/interpreter/VoiceInterpreter.java
```diff
@@ -1147,24 +1147,53 @@ private void conferenceStateModeratorPresent(final Object message) {
...
+                    logger.debug("%%%%%%%%%%% Identified customer header: " + headerName);
                 }
                 parameters.add(new BasicNameValuePair(prefix + headerName, sipMessage.getHeader(headerName)));
+            } else if (headerName.startsWith("Diversion")) {
+
+                final String sipDiversionHeader = sipMessage.getHeader(headerName);
+                if (logger.isDebugEnabled()) {
+                    logger.debug("%%%%%%%%%%% Identified diversion header: " + sipDiversionHeader);
+                }
+                parameters.add(new BasicNameValuePair(prefix + headerName, sipDiversionHeader));
```
antonmry: @otsakir , I have use this nice feature where the custom headers are added with a prefix, in this case, the Header is "diversion", so that's the RVD variable

antonmry:
@otsakir , not sure if you are notified with line notes, check my comment, please.

otsakir:
@antonmry, indeed the direction SipHeader_direction is forwarded to RVD nicely as you describe. The behaviour was not really expected since it relies on the fact that we don't do much of sanitization for SipHeader_* paramaters from the RVD side.

Anyway, let's not bother now about it. It's clear from my side (RVD). @gvagenas , please proceed to reviewing.

antonmry:
@otsakir, yes, I agree it's not obvious. I did it in that way because that variable is more for debugging and it avoided to have "not important" variables in RVD, as it's the case with custom headers.

otsakir:
It makes sense. Thanks for contributing this  @antonmry ;-)

PR #1253, opened 2016-07-02
Issue1020_RefactorAutoConfigScripts

ghjansen:
Some considerations:
- The autoconfigure scripts are now refactored to work with `mediaserver.xml` file instead of `server-beans.xml` file;
- The configurable variables related to RMS inside `restcomm.conf` remains the same as before;
- I spinned RMS version inside `build.xml` to work with version 4.2.0 build 72 which, so the result from building from source is a operational build given the new configuration mechanism;
- The application `sed` that we use in the configuration scripts appears to have two different versions, one for GNU/Linux and other for BSD/OSX what seems to bring incompatibility sometimes, like described here: http://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux , so:
  - I faced multiple error messages from `sed` in my environment (BSD/OSX) like is shown by file 
    [start-restcomm-osx-problematic.txt](https://github.com/RestComm/Restcomm-Connect/files/344427/start-restcomm-osx-problematic.txt)
    and applied the quick fix of including the extension parameter when `sed` uses the in-place flag, once this parameter is mandatory in BSD/OSX. This way, some lines like `sed -i` are now `sed -i ''`. After this adjustment there is no longer error messages, as shown by file 
    [start-restcomm-osx-solved.txt](https://github.com/RestComm/Restcomm-Connect/files/344428/start-restcomm-osx-solved.txt);
  - Considering that the issue is related to the environment, i performed some tests on GNU/Linux as well after the adjustments to check if some conflict happens. I got multiple error messages from `sed` but im almost sure that this can be a problem with my environment variables, since i just created a VM and only set RESTCOMM_HOME. The result is shown by the file 
    [start-restcomm-linux-problematic.txt](https://github.com/RestComm/Restcomm-Connect/files/344429/start-restcomm-linux-problematic.txt)
  - Given that, i just kept the BSD/OSX adjustments so we can just discard those changes if you guys may see them as a thread for GNU/Linux side, i skipped further tests on GNU/Linux because otherwise i wouldn't finish this in time.

Please, if possible, let me know until Sunday if there are adjustments that you may see as necessary to this patch, otherwise, feel free to discard part of the work if its irregular or for any other reasons.


restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-rms.sh
```diff
@@ -14,26 +14,25 @@ configServerBeans() {
...
+	sed -e "s|<bindAddress>.*<\/bindAddress>|<bindAddress>$1<\/bindAddress>|" \
+		-e "s|<externalAddress>.*</externalAddress>|<externalAddress>$MSERVER_EXTERNAL_ADDRESS</externalAddress>|" \
+	    -e "s|<network>.*<\/network>|<network>$2<\/network>|" \
+	    -e "s|<subnet>.*<\/subnet>|<subnet>$3<\/subnet>|" \
+	    -e "s|<sbc>.*<\/sbc>|<sbc>$USESBC<\/sbc>|" \
+	    -e "s|<address>.*<\/address>|<address>$1<\/address>|" \
+    	-e "s|<port>.*<\/port>|<port>${REMOTEMGCP}<\/port>|" \
+	    -e "s|<lowPort>.*<\/lowPort>|<lowPort>$MEDIASERVER_LOWEST_PORT<\/lowPort>|" \
+	    -e "s|<highPort>.*<\/highPort>|<highPort>$MEDIASERVER_HIGHEST_PORT<\/highPort>|" \
+	    -e "s|<dtmfDetector poolSize=\"\(.*\)\" dbi=\"\(.*\)\" \/>|<dtmfDetector poolSize=\"\1\" dbi=\"$DTMFDBI\" \/>|" \
```
hrosa: The DTMF detector pool size should be configurable, not statically set to 1

ghjansen:
This PR seems obsolete and should be closed without merging, since @leftyb apparently took care of few problems i found while working there, through issue #1267 i guess. Few of the issues found and also be avoided today by using the [gsed script](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/gsed_macos.sh) for OSX.

Related to #2447.

Issue #1168, opened 2016-06-09
Support account role update


otsakir:
Account role update is not supported. The existing policy used in account *creation* is the following:

* Administrators can explicitly set the roles of the sub-accounts they create.
* non-administrators create sub-accounts of  their own role.

The same policy can be adapted to account modification:

* Administrators can explicitly update the role of a sub-account arbitrarily.
* non-administrator cannot change the role of their sub-accounts at all.
* Administrators can change their own role arbitrarily too. That may imply privilege loss and a warning can be displayed in the UI in such a case.

The patch includes patching AccountsEndpoint, AdminUI account details form and the AccountsEndpointTest in the testsuite (check role is updated).

muhammadbilal19:
@otsakir i stated making some changes for this issue. I need to clarify few things

- if Current login user Say Default Admin (admin@company.com) want to update own role do we need to allow ? In that case do system need to force this user to login again?
- According to the rule non-administrator can only create sub accounts of their own types but in that scenario their is a possibility that Default Admin may have few administrator sub accounts

muhammadbilal19:
@otsakir can you please comment on my previous comment.

otsakir:
@muhammadbilal19, appologies for the delay. Your comment fell through the cracks :-(.

* Indeed, it seems that allowing an admin to change his own role is not really necessary, i agree.
* The Default Admin may have administrator sub-accounts. I think that's ok. Default Admin should take care when someone is made admin (not sure if i got your second bullet here). 
* In general, allowing an account changing his own role seems like a useless feature.

Thinking over it, it seems to me that the only case of role modification that is really needed is for an Admin to be allowed to change the roles of his sub-accounts arbitrarily.

@muhammadbilal19 wdyt?
@gvagenas, @deruelle , any thoughts on this ?

deruelle:
As a general rule, there is some documentation on accounts and sub accounts at http://documentation.telestax.com/connect/api/Restcomm%20-%20Multi-tenancy%20and%20Managing%20Sub-Accounts.html#restcomm-rest-api. If there is not enough information. Please consider contributing changes to it at https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api/Restcomm%20-%20Multi-tenancy%20and%20Managing%20Sub-Accounts.adoc

1. "allowing an admin to change his own role is not really necessary" i agree.
[...]

PR #1294, opened 2016-07-20
Issue 1168

muhammadbilal19:
Please review my changes for the account update issue #1168. Feel free to contact me in case you see any issue.


deruelle:
Thanks @muhammadbilal19 

@otsakir can you review as time permits ?

otsakir:
Sure @deruelle.

otsakir:
@muhammadbilal19, i've merged your PR! 

I've also enforced access control rules on account role update at the restcomm side. So that the only case role-update is permitted using the REST api directly is from administrator when they are setting roles on their sub-accounts.

Thanks for contributing this patch :+1:

PR #1298, opened 2016-07-25
Fix Usage endpoint (call duration) for  MariaDB

### LINKED ISSUE #309: Dashboard call analytics - orange minutes line and right axis make no sense
1. The minutes are always 0 even though the call count varies
2. The minutes axis has negative numbers which makes no sense for counting minutes.

screen shot 2015-06-23 at 12 53 43 am

brainslog:
Implement changes to have the Usage endpoint and namely the duration
field for calls to work properly both at HSQLDB and MariaDB.

Needed to break the query where the column name goes as MariaDB won't
accept " in column name and HSQLDB requires it. In the proper usage.xml
it is being added or not, as required.

Fixes #309


deruelle:
Thanks @brainslog. @gvagenas @maria-farooq can you please review ?

gvagenas:
Thanks @brainslog . Patch reviewed and merged to master

PR #1560, opened 2016-11-10
Support for RMS 5.x auto-configuration

hrosa:


leftyb:
@hrosa @gvagenas I think some basic description of the configuration options at mediaserver.conf
needs to be added at the file it self. Although the documentation is updated [doc](http://documentation.telestax.com/core/media_server/Media_Server_User_Guide.html#_ctms_configuring_the_media_server)

leftyb:
Media server log directory is not created (`/Restcomm-JBoss-***/mediaserver/log`) when starting  restcomm from `start-restcomm.sh`

leftyb:
Error when enabling https on RMS. 
need to add ```JAVA_OPTS_TRUSTORE="-Djavax.net.ssl.trustStore=$CERTIFICATION_FILE -Djavax.net.ssl.trustStorePassword=$TRUSTSTORE_PASSWORD"``` at the RMS java_opts.
Based at  [this](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-SecureSSL.sh#L100)

leftyb:
Got distortion voice for 1234 demo call.
Solution:
need to change the RMS codecs to: ```MEDIA_CODECS=pcmu,pcma,telephone-event```

maria-farooq:
@gvagenas After testing RMS5 has green light and we can merge this PR.

gvagenas:
@hrosa @leftyb the new configuration scripts should be simplified to follow the default standards we used to have and more important the IP Address configuration. 

Currently, if we provide ip address details for Restcomm, the same details will be used for media server UNLESS we provide specific ip address details for media server. We should follow the same for the RMS-5 configuration.

I think we should not have a new mediaserver.conf at the auto-config scripts. If we need to edit media server configuration, why not edit the media server conf file directly? 
We should try to reduce configuration files and options, not introduce new.

Also, I don't think we should have the start/stop-mediaserver.sh scripts, no need to have them there also.

Lets keep in mind that the purpose of the auto-config script is to help users work with Restcomm  by providing sensitive defaults and a set of basic configuration options that they might need, without the need to go over all the configuration files etc.

leftyb:
@gvagenas @maria-farooq @hrosa  The start & stop scripts for RMS are actually used at the start-restcomm.sh script. (edited)
So I dont see any issue using specific scripts for it.

As as well permits the separation between the two Application level concepts. (edited)
For docker usage as well are pretty handy. We can use the to create separate container and as well to launch them separately from the  application supervision perspective..

Finally I dont think so bad to have all configuration files at one place (.conf) (bin/restcomm). It is much better from user perspective to be able to configure RC from the same place than CDing to different directories.

PR #1561, opened 2016-11-10
Issue1392 Add global notification to Admin UI logs/notification

croufay:
Issue1392 Add global notification to Admin UI logs/notification

First iteration is complete. 

TODO

update the http://documentation.telestax.com/connect/api/error-reference.html
after adding more notification

### DIFF
--- a/restcomm/restcomm.notification/src/main/java/org/restcomm/connect/notification/GlobalNotification.java
+++ b/restcomm/restcomm.notification/src/main/java/org/restcomm/connect/notification/GlobalNotification.java
@@ -1,0 +1,95 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.restcomm.connect.notification;
+
+import java.net.URI;
+import org.apache.commons.configuration.Configuration;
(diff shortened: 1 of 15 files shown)


restcomm/restcomm.notification/src/main/java/org/restcomm/connect/notification/GlobalNotification.java
```diff
@@ -0,0 +1,96 @@
...
+import org.restcomm.connect.dao.DaoManager;
+import org.restcomm.connect.dao.NotificationsDao;
+import org.restcomm.connect.dao.entities.Notification;
+import org.restcomm.connect.commons.dao.Sid;
+
+/**
+ *
+ * @author charles.roufay
+ */
+public class GlobalNotification {
```
gvagenas: The GlobalNotification should be an Akka actor that the rest of the actors should use to send notification to be created.
For example, from CallManager we should have notificationService.tell("notification blabla", self());
Also maybe provide a Notification API so we can have different implementations of the Notification service, for example one implementation would only generate Notification another implementation would also send SMS for certain types of Notifications.

gsaslis:
Hey @croufay!
doing some housekeeping here... is this still a valid PR ?

PR #1598, opened 2016-11-23
Issue1497 strong password hash

otsakir:
Adds bcrypt hashing algorithm to passwords


gsaslis:
@otsakir / @gvagenas doing some housekeeping here... is this still a valid PR ?

otsakir:
As with 1496 PR i don't think it's valid @gsaslis . But please check with @gvagenas or @maria-farooq if they have any use of it.

gsaslis:
Great stuff @otsakir. thanks!

Closing so we can clean up a bit - @gvagenas / @maria-farooq plz do feel free to reopen if need be.

PR #1603, opened 2016-11-23
Trustcdr cdr logger

maria-farooq:
# 1562


maria-farooq:
we need to re-discuss design and then move forward on it. until then it can be considered blocked.

Issue #1596, opened 2016-11-22
Application permissions should affect access to sub-account resources too


otsakir:
Currently the Restcomm:Applications permission only controls access to the applications of the operating account (the account specified in credentials). Access to the children of this account is forbidden by default. 

Instead of forbidding we should allow the same type of access as the one that applies to the operating account.

otsakir:
issue is fixed

PR #1604, opened 2016-11-24
Issue1596 review application access rules and optimize account closing

otsakir:
The patch does the following:
1. Relaxes application REST API access rules (allows parent accounts access sub-account applications) - #1596 
2. Reviews RVD notification mechanism for account removal from RVD side. RVD now queries Restcomm to get application for a specific (removed) account instead of going through all the project in the workspace
3. Refactors account closing so that it sends one notification per account removed instead of a single notification array. The reason is to reduce complexity in the account closing implementation.


otsakir:
There is also this ready-to-use binary:
https://mobicents.ci.cloudbees.com/view/RestComm/job/RestComm-For-Specific-Branch/13/

Issue #1568, opened 2016-11-14
Using a bad account SID in Call creation API returns stack trace


otsakir:
Instead just a 500 response with a small error description should be returned.

forever3000:
Hi @otsakir ,
As @deruelle 's assign, I will try to handle this issue. Can you give me more details.
Thanks and Best Regards,
Vu Nguyen

otsakir:
Hi @forever3000.

Thanks for taking the step and getting your hands dirty! ;-)

Well, i would start with registering a client, creating a call to this client using the Calls REST api like the following:

http://documentation.telestax.com/connect/api/calls-api.html#making-a-call-to-a-restcomm-client

And then replacing the Account SID in the url with (a) a totally broken account SID (like 'crappy-sid') or (b) a _syntactically valid but not existing_ SID like AC00000000000000000000000000000000.

That should do the trick.

Ping here if you need any help.

forever3000:
Hi @otsakir ,

Sorry for bothering you more :). This's my very first time that I'm working on Restcomm-Connect so can not speed up. 
After start Restcomm-Connect server, I tested by making call between Alice and Bob on Jitsi SIP Phone to make sure server working properly.
Then I tried to make a call to SIP account as below:
`curl -X POST https://ACae6e420f425248d6a26948c17a9e2acf:5471175aea6081a114cc5c9e5bb99bae@10.211.55.4:8080/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Calls.json -d "From=+16175551212" -d "To=sip:alice@10.211.55.4:8080" -d "Url=http://10.211.55.4:8080/restcomm/demos/hello-play.xml"`

As I understand 
ACae6e420f425248d6a26948c17a9e2acf: SID of admin account
5471175aea6081a114cc5c9e5bb99bae: maybe the auth token
10.211.55.4: my private IP
2012-04-24: I'm not sure about this one
And I always got : curl: (35) Unknown SSL protocol error in connection to 10.211.55.4:8080

May you have some advices for me. Beside that, can you give me how to register a client with curl command :).

Thanks and Best Regards,
Vu Nguyen

otsakir:
Hi forever3000,

1. In local installations restcomm runs on plain http by default. So, you have to access the REST API under http:// and not https://.
2. ACae6e420f425248d6a26948c17a9e2acf is the SID of the account
[...]

PR #1619, opened 2016-12-04
Fix Using a bad account SID in Call creation API returns stack trace … issue #1568

forever3000:
…#1568 issue


deruelle:
Thanks @forever3000. @otsakir can you review ?

otsakir:
sure @deruelle

PR #1673, opened 2016-12-21
Improvements on Performance Jobs

hrosa:


gsaslis:
@hrosa / @gvagenas i take it this is no longer relevant?

Should I close? (reply, or else... ;) )

Issue #1690, opened 2017-01-04
Usage API the start date is not respected


deruelle:
See below, start date is 1st December but the records starts at 2nd

curl --user 'ACXXX:YYYYY' 'https://cloud.restcomm.com/restcomm/2012-04-24/Accounts/ACXXXX/Usage/Records/Monthly.json?EndDate=2016-12-31&StartDate=2016-12-01&Category=CALLS'
[
{
"category": "calls",
"description": "Total Calls",
"account_sid": "ACXXX",
"start_date": "2016-12-02",
"end_date": "2016-12-31",
"usage": "12402",
"usage_unit": "minutes",
"count": "638",
"count_unit": "calls",
"price": "0",
"price_unit": "USD",
"uri": "/todo.json"
}

True for both Daily and Monthly

abdulazizali77:
So what wasnt immediately clear to me was (given all prior incorrect guesses) this is a simple aggregate query that just displays available records from MIN(date_created) to MAX(date_created) within the started_date and end_date bounds. We see 2016-12-02 as start_date only because *i assume* there are no records on 2016-12-02, and miraculously there are records on 2016-12-31 which is why end_date is correct, else it wouldve also been incorrect.
Potentially this is correct behaviour, another alternative fix could be to introduce two extra API attributes to display available lower and upper bound records.
PR #1698 however only replaces the current start_date and end_date with the passed in parameters

deruelle:
@abdulazizali77 @otsakir can we close this issue ?

PR #1701, opened 2017-01-10
Issue #1690: change display start_date and end_date 

abdulazizali77:
Previously it was returning the date_created column as the values of start_date and end_date. Now we display those as the passed in parameters instead.

### DIFF
--- a/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/usage.xml
+++ b/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/usage.xml
@@ -13,8 +13,8 @@
 			COUNT(1) AS "count",
 			CAST(${usageExprPre}${usageExprCol}${usageExprSuf} AS SIGNED) AS "usage",
 			SUM(CONVERT("price", SIGNED)) AS "price",
-			CONVERT(MIN(date_created), DATE) AS "start_date",
-			CONVERT(MAX(date_created), DATE) AS "end_date",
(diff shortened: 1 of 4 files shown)


abdulazizali77:
@otsakir  @deruelle Please review thanks.

deruelle:
@abdulazizali77 Thanks for the pull request. @otsakir will review. One question, can you actually verify that the usage and usage_unit is correct, I have a hunch that we display a value in minutes that should actually be seconds. It seems the value shown by minutes is too high

abdulazizali77:
@deruelle Youre right, its actually seconds, i overlooked it. Should i fix it along with this bug? 
I reckon it should just be a change in the select query

deruelle:
@abdulazizali77 yes please, thanks !

abdulazizali77:
@deruelle Should we display the minutes usage as
i) float ? eg 11.60, 18.01
ii) integer rounded up down? eg 12, 18

iii) I thought the best (and hardest) wouldve been free form text, "x minutes y seconds"....
one way to approximate this potentially would be:
<pre>
"usage": "206, 42",
"usage_unit:" "minutes, seconds"
</pre>
what do you think?

deruelle:
@abdulazizali77 let's go with integer rounded up. this usually used for billing reasons that are billed by the minute consumed

abdulazizali77:
@deruelle done!

abdulazizali77:
The sql changes in PR#1696 will address any Start/EndDate query param issues. The minutes usage fix is also ported to #1696. So i am closing this

Issue #1712, opened 2017-01-13
Account API response "uri" property inconsistent with actual uri


otsakir:
The RESTfull response when retrieving an account has a minor inconsistency in the _uri_ field. It returns:

```
"uri": "/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf.json",
```

while the endpoint expects:
```
"uri": "/2012-04-24/Accounts.json/ACae6e420f425248d6a26948c17a9e2acf",
```

It comes down to the fact that the Accounts REST API is defined a little differently from other APIs.

abdulazizali77:
ill take it

deruelle:
@abdulazizali77 great ! assigned to you

deruelle:
@abdulazizali77 would you be ready to take on #898 as well to fix our API inconsistencies in that regard ?

abdulazizali77:
@deruelle I will start looking into it!

otsakir:
@abdulazizali77, @deruelle, i went through the PR and also ran a couple of manual tests to see the responses directly. There is still not quite right.

I requested my account:

`curl 'http://this:8080/restcomm/2012-04-24/Accounts.json/ACae6e420f425248d6a26948c17a9e2acf'`

and got this response
```
{
  "sid": "ACae6e420f425248d6a26948c17a9e2acf",
  "friendly_name": "Default Administrator Account",
  "email_address": "administrator@company.com",
  "type": "Full",
  "status": "active",
  "role": "Administrator",
  "date_created": "Tue, 24 Apr 2012 00:00:00 +0300",
  "date_updated": "Mon, 16 Jan 2017 12:31:02 +0200",
  "auth_token": ...
  "uri": "/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf",
  "subresource_uris": {
    "available_phone_numbers": "/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/AvailablePhoneNumbers.json",
    "calls": "/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Calls.json",
    "conferences": "/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Conferences.json",
    "incoming_phone_numbers": "/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/IncomingPhoneNumbers.json",
[...]

Issue #898, opened 2016-02-25
Inconsistencies in Accounts* REST Endpoint 


otsakir:
Restcomm offers two set's of methods for accessing Account resources as JSON or XML. Here are the *current* URLs for all methods available:

#### AccountsJsonEndpoint.java

1 - **retrieve all** (JSON)
GET /Accounts.json

2 - **retrieve one** (JSON)
GET /Accounts.json/{accountSid}

3 - **delete**
DELETE /Accounts.json/{accountSid}.json

4 - **create** (JSON)
POST /Accounts.json

#### AccountsXmlEndpoint.java

5 - **retrieve all** (XML)
GET /Accounts

6 - **retrive one** (XML)
GET /Accounts/{accountSid}

7 - **delete** (XML)
DELETE /Accounts/{accountSid}

8 - **create** (XML)
POST /Accounts

9 - **update** (JSON)
POST /Accounts/{accountSid}.json
PUT /Accounts/{accountSid}.json

10 - **update** (XML)
POST /Accounts/{accountSid}
PUT /Accounts/{accountSid}

#### Suggestions

* The separation into AccountsXmlEndpoint and AccountsJSONEndpoint doesn't make sense and is confusing since the methods are mixed up. We should either split methods correctly or if JAX-RS doesn't permit this put them all in a single class.
* **.json** suffix should be placed in a single place, at the end. In that sence (2) and (3) are wrong. Twilio puts the suffix at the end:
https://www.twilio.com/docs/api/rest/account

Here is how the propsed endpoint would be like:

#### AccountsRestEndpoint.java:
Path: */Accounts*

1 - **retrieve all** (JSON)
GET /Accounts.json

2 - **retrieve one** (JSON)
GET /Accounts/{accountSid}.json

3 - **delete** (JSON)
DELETE /Accounts/{accountSid}.json

4 - **create** (JSON)
POST /Accounts.json

5 - **update** (JSON)
POST /Accounts/{accountSid}.json
PUT /Accounts/{accountSid}.json

6 - **retrieve all** (XML)
GET /Accounts

7 - **retrive one** (XML)
GET /Accounts/{accountSid}

8 - **delete** (XML)
DELETE /Accounts/{accountSid}

9 - **create** (XML)
POST /Accounts

10 - **update** (XML)
POST /Accounts/{accountSid}
PUT /Accounts/{accountSid}

@gvagenas thoughts ?

ghjansen:
[...]

PR #1715, opened 2017-01-14
Issue #1712: Accounts API - dont return .json suffix for Uri resource

abdulazizali77:
This is also related to #898 
@otsakir @deruelle This PR might not be applicable if we *change our minds*

### DIFF
--- a/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/converter/AccountConverter.java
+++ b/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/converter/AccountConverter.java
@@ -20,7 +20,6 @@
 package org.restcomm.connect.http.converter;
 
 import java.lang.reflect.Type;
-import java.net.URI;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
--- a/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/converter/AccountConverter.java
+++ b/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/converter/AccountConverter.java
@@ -20,12 +20,12 @@
 package org.restcomm.connect.http.converter;
 
 import java.lang.reflect.Type;
+import java.net.URI;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonSerializationContext;
 import com.google.gson.JsonSerializer;
-
 import com.thoughtworks.xstream.converters.MarshallingContext;
 import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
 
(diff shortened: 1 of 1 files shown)


deruelle:
Thanks @abdulazizali77. @otsakir please review

deruelle:
@abdulazizali77 we may change our mind as part of #898  review but at least the uri is consistent with the API call that was made. 
#898 should solve the inconsistencies across the board but it may take a bit of time to fix them all so I'm fine with moving forward on consistency iteratively

abdulazizali77:
@otsakir please rereview
An alternative fix without modifying method signatures wouldve been to check against request.getCintentType() == 'application/json', bit kludgy but wouldnt have to pass an extra param, tell me what you think.

otsakir:
@abdulazizali77, i think it will work for now. It's better to put an exception to the rule in the upper layers (endpoint) like you did than hide it in the converter. Let me get on with the merge..

otsakir:
Hmmm, @abdulazizali77, the concept does not work. `createFrom()` is not involved when returning accounts. Only when creating new ones. The initial approach seems better after all.

In AccountConverter.java, i would overload prefix() so that it takes an additional content type parameter and use this overloated version from `writeUri(final URI uri, final JsonObject object) ` which i _know_ works for JSON content type. wdyt ?

abdulazizali77:
@otsakir I agree with your idea but i would rather use MediaType like everywhere else
<pre>private String prefix(final Account account, final MediaType responseType) </pre>
What do you think?
This would also mean that we wouldnt be using the persisted uri at all in returning the response! We will address that issue later
I realize now that my prior implementation wouldve mapped a specific Account ID to either XML or JSON uri, when it shouldnt actually be exclusively either!

otsakir:
> @otsakir I agree with your idea but i would rather use MediaType like everywhere else

ok @abdulazizali77, sounds great. 

> This would also mean that we wouldnt be using the persisted uri at all in returning the response!

Hmm, indeed. Wondering whether we should involve the data layer or not here. Btw, there has also been a suggestion in the past to get rid of content-type suffixes (.json etc.) and use the respective http headers instead. What is your opinion on that ?

> I realize now that my prior implementation wouldve mapped a specific Account ID to either XML or JSON uri, when it shouldnt actually be exclusively either!

well, that's what PRs are for. A second pair of eyes never harmed anyone ;-)

abdulazizali77:
> well, that's what PRs are for. A second pair of eyes never harmed anyone ;-)
Ha, im sure everyone has had 'extreme programming' with three engineers and one manager behind them.... :D

abdulazizali77:
>Btw, there has also been a suggestion in the past to get rid of content-type suffixes (.json etc.) and use the respective http headers instead. What is your opinion on that ?
String/URI rewriting is a bad thing i feel, but i think we should support and standardize what we have right now for this version and add content type header based webservices in the next version on top of the hopefully deprecated suffix based one

abdulazizali77:
@otsakir ready

gvagenas:
Thanks @abdulazizali77. Thanks @otsakir for review. PR is merged

PR #1727, opened 2017-01-18
Fix to send BYE for caller when all branches are BUSY

ghjansen:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -1101,7 +1101,13 @@ private void onCallStateChanged(Object message, ActorRef sender) throws Transiti
                         if (dialBranches != null && dialBranches.contains(sender)) {
                             removeDialBranch(message, sender);
                         }
-                        checkDialBranch(message, sender, attribute);
+                        if (dialBranches != null && !dialBranches.isEmpty()){
```
gvagenas: @ghjansen this shouldn't be removed. The intention here is the following:
1. Every BUSY branch will be removed from dialBranches
2. If dialBranches is empty, the checkDialBranch() method will ask for the next verb
3. If next verb is END, Bye will be sent to the call

If you cannot see the above, then maybe there is a race condition that moves the call to finishDialing, but there also the Call should get a Bye.

ghjansen: I see, thanks @gvagenas ! But i understood from a previous conversation on slack with you and @deruelle  that if **all** or **the only** dial branch are/is busy, the interpreter should not proceed with the RCML and disconnecting the call instead. Please check [this message](https://telestax.slack.com/archives/restcomm-connect/p1482484549001229). Because of that, i assumed that `checkDialBranch` should be performed only if `dialBranches` is not empty, otherwise the call should be finished.

But, if i got it wrong and it has to be the way you mentioned now by steps 1, 2 and 3, i suppose we'll face that exception about an invalid transition from `forking` state to `checkingCache` state, since the flow goes from a Dial verb receiving busy to a Play verb.

I wait for your consideration on that to continue. Thanks! :)

gvagenas:
@ghjansen Seems that at slack there was a misunderstanding. Here is the proper call flow for dial fork with all branches busy:
https://www.websequencediagrams.com/cgi-bin/cdraw?lz=cGFydGljaXBhbnQgQWxpY2UKAAYMUmVzdGNvbW0ABw5DTUxfQXBwAB0NQm9iAC0NQ2Fyb2wAPw1EYW4KCgBeBS0-AFAIOiBJbnZpdGUgdG8gRElEIHRoYXQgcnVucyBEaWFsIEZvcmsgYXBwCgCBAAgtPgCBHwU6IDEwMC8xODAKAA8LAIEPCDogR2UAgR8GCgCBIggAZQxGaXJzdCB2ZXJiIG9mAIFHBSAoRElBTCkAQgogLT4gQm9iAIEVCAALDQCBUwUABRVEYW4AKQkKQm9iIC0-AII3CQCBLAUgVHJ5aW5nAAoSNDg2IEJ1c3kgaGVyZQCBRg0Agh4JaWYgKERpYWxCcmFuY2hlcyAhPSBudWxsKSB7XG4gcmVtb3ZlIGN1cnJlbnQgYgAgBSBhbmRcbiB3YWl0IGZvciBvdGhlcgAVB2VzIHJlc3BvbnNlXG59IGVsc2UARAVjaGVjawApBW5leACCHwdpbgCCIAZBUFB9IAoKbm90ZSByaWdodCBvZgCELQY6IFRoZXJlIGFyZSBtb3JlIGRpYWwAYAlwZW5kaW5nAIEiBQBtBSBzbwCEUAkAgRoFcwoKAIQqBQCCGRkACxMAgQqBKgCBW1hEYW4AhEoZAAsRAG6BQU5vAIRJEiwAhRwUAIc5DgCIBw4AGwoAiAYURW5kIChlbmQAiBAIAIgHCwCIbwlCeWUK&s=rose

Also check the test case: org.restcomm.connect.testsuite.telephony.DialForkTest#testDialForkEveryoneBusyExecuteRCML_ReturnedFromActionURL where all branches are busy and after Dial verb there are more RCML verbs to be executed.

George

ghjansen:
Closed without merging as we discovered this patch is not valid. The issue was properly fixed at #2271 (#2432 + PR #2420) turning this PR into obsolete.

PR #1760, opened 2017-02-03
No play actions in 'waitingForAnswer' state

kacmos:


kacmos:
@gvagenas 
Please, check new junit test

kacmos:
@gvagenas 
take a look on the new tests case in ClientsDialAnswerDelayRingingErrorTest.
It illustrates the failure that occurs without issue 1759 patch

PR #1775, opened 2017-02-07
update script to prevent multiple entries in /etc/hosts file

croufay:
update script to prevent multiple entries in /etc/hosts file


leftyb:
@croufay The script as it is should add the host only if not exist. It does not work as it is?

croufay:
@leftyb this doesn't work as expected that is why I updated the script and the customer conformed.

gvagenas:
@leftyb @croufay what is the status of this PR?

croufay:
@leftyb please look into this as we need this to be included in the next update for a customer

leftyb:
@croufay the problem with your solution is that if a public domain is used then it will be overwritten by the resolution set at /etc/hosts. I just have tested the script as it is and works properly.

First checks if the domain is a public domain (which means that is no need to add at /etc/hosts), and it adds it at the /etc/hosts only if it is not a public domain.

You need to explain under which use case, the current script is not working.

Regards.

PR #1788, opened 2017-02-09
Registration removal when IMS answers with error.

kacmos:
testReRegisterClientForbidden in ImsClientsDialTest is the regression test


gvagenas:
Thanks @kacmos , PR review finished and merged to master

Issue #1738, opened 2017-01-26
RESTfull response inconsistencies - II


otsakir:
The following request reveal mostly minor inconsistencies wrt to the uri/subresource_uris fields returned:

1. http://this:8080/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/IncomingPhoneNumbers.json

    Response still contains the 'restcomm' part:

    "uri": "/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/IncomingPhoneNumbers/PNdd7a0a0248244615978bd5781598e5eb.json"

2. This is dead (returns 404). There is no REST method for this path. 
http://this:8080/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/AvailablePhoneNumbers.json

3. Calls list method has no .json suffix in 'subresource_uris'
test uri: 
    http://this:8080/restcomm/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Calls.json
response:
    ...
      "subresource_uris": {
        "notifications": "/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Calls/CA1cbea362aaef4482a941af600f3f7f05/Notifications",
        "recordings": "/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Calls/CA1cbea362aaef4482a941af600f3f7f05/Recordings"
    ...
    
4. Errors /2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Conferences.json
 /2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Conferences.json

    a. 'last_page_uri' looks like: "last_page_uri": "/2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Conferences.json?Page\u003d0\u0026PageSize\u003d50" ( '=', '&' are encoded)
    b. 'conferences.participants' looks like: /2012-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/Conferences/CF1e4a2e67ada54298a83b93818c0ea1e4/Participants (no .json extension)
    c. 'subresource_uris' is empty - {} 

This issue extends fixes started in #1723 .

abdulazizali77:
@otsakir I guess i should look at these! :D
Potentially this should be linked to #898

[...]

PR #1801, opened 2017-02-12
Issue #1738 - Various EP fixes

abdulazizali77:
@otsakir Apologies for the slow update. This fixes 1, 3, 4. 
for 4a we could call `gson.disableHtmlEscaping()` in all EPs, but i did not do that in this PR.

I will address 2) in a separate PR


deruelle:
@abdulazizali77 thanks. @otsakir can you review ?

otsakir:
@abdulazizali77 sorry for the delay in merging. It has been for a good reason.

A discussion with the team has lead to some interesting conclusion. The fact that we store 'uri' in the DB is causing more trouble that gain. Why not produce this dynamically ? Maybe create a central mechanism, or customize the converters more. All required bits of information to produce the url are already there. Also in terms of performance, there is not much more involved than string concat which is already the case.

Also see https://github.com/RestComm/Restcomm-Connect/issues/1976 .

otsakir:
@abdulazizali77 , after discussing with @gvagenas it seems that you need to strip the '/restcomm/' part from mariadb init scripts too (init.sql). 

@gvagenas can you confirm that ?

abdulazizali77:
sorry i missed it! @otsakir @gvagenas, c'est done!

PR #1806, opened 2017-02-12
Issue#708 Add PhoneNumber Regex for SMS/Voice and USSD

croufay:
Issue#708 Add PhoneNumber Regex for SMS/Voice and Ussd

The code does a normal exact match and if that fails, it checks if there is a relevant Regex match  

example regex 

tested

^*77...33#$ = matches any number that starts with *77 and and the three dots  matches any character and it must end with a # sign

^[12]2233#$ =  matches any number that starts with 1 or 2 and ends with 2233#

^[*]76[*]88[456]#$ =  matches any number that starts with *76*884# or *76*886# or *76*886#

7777|8888 = matches 7777 or 8888

[45]234[23] =  matches 42342, 42343,   52342 and 52343,  

All tested with Voice, SMS or USSD where relevant.


croufay:
I;m assigning to @maria-farooq

maria-farooq:
Hi @croufay i reviewed other 2 PRs 
- https://github.com/RestComm/Restcomm-Connect/pull/1896
 and
- https://github.com/RestComm/Restcomm-Connect/pull/1927 

This one i think should be reviewed by @gvagenas 

Thanks
Maria

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
```diff
@@ -70,19 +74,61 @@ public IncomingPhoneNumber getIncomingPhoneNumber(final String phoneNumber) {
...
-            if (result != null) {
+            if (result != null ) {
                 return toIncomingPhoneNumber(result);
-            } else {
-                return null;
-            }
-        } finally {
+            }//}
+            //check if there is a Regex match only if parameter is a String aka phone Number
+            if(!(parameter instanceof Sid)){
```
gvagenas: @croufay why you check if object parameter is not an instance of Sid? What is the point here?

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
```diff
@@ -70,19 +74,61 @@ public IncomingPhoneNumber getIncomingPhoneNumber(final String phoneNumber) {
...
+               inboundPhoneNumber = parameter.toString().replace("+1", "");
```
gvagenas: @croufay why you always remove the +1 sign? What if we have `+1512*`, what if we have `+331554*`?
The `+` sign should be removed after we check the DB and there is no match

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
```diff
@@ -70,19 +74,61 @@ public IncomingPhoneNumber getIncomingPhoneNumber(final String phoneNumber) {
...
+               logger.warn(" Inbound phoneNumber : " + inboundPhoneNumber );
```
gvagenas: @croufay no need for WARN log here. Just make it INFO

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
```diff
@@ -70,19 +74,61 @@ public IncomingPhoneNumber getIncomingPhoneNumber(final String phoneNumber) {
...
+          return null;
+
+}
+
+   public IncomingPhoneNumber checkIncomingPhoneNumberRegexMatch ( String selector, String inBoundPhoneNumber){
+               final SqlSession session = sessions.openSession();
+               String phoneRegexPattern = null;
+               try {
+                    List<IncomingPhoneNumber> listPhones = getAllIncomingPhoneNumbers();
+                   for (IncomingPhoneNumber listPhone : listPhones){
```
gvagenas: @croufay I don't get the reason to go through all the stored numbers to build the Regex pattern. Can you elaborate? This doesn't seem to be an efficient way.

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
croufay: @gvagenas, that shouldn't be a problem because the first part always checks for exact match. ex. +331554*, If there is no exact match, it does a regex match. You have to remove the + sign as it has a special meaning and it will cause the regex search to fail

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
croufay: OK, I can update this to INFO

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
croufay: @gvagenas, that is true but I didn't want to mess with the exact match we already have in place. Another thing is that the exact match is done using predefined SQL that we have defined in the https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.application/src/main/webapp/WEB-INF/sql/incoming-phone-numbers.xml

The other option would be the parse the Regex using mysql but that might be problematic if mysql or mariadb version changed as we have seen when we do DB upgrade.

croufay:
@gvagenas , the test case are similar to what we already do to match any numbers. We just need to add a few of the scenarios below

^*77...33#$ = matches any number that starts with *77 and and the three dots matches any character and it must end with a # sign

^[12]2233#$ = matches any number that starts with 1 or 2 and ends with 2233#

^[]76[]88[456]#$ = matches any number that starts with 76884# or 76886# or 76886#

7777|8888 = matches 7777 or 8888

[45]234[23] = matches 42342, 42343, 52342 and 52343,

All tested with Voice, SMS or USSD where relevant.

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
gvagenas: @croufay here you don't remove just the `+` sign but the `+1` so I guess you have to fix that, right? 
Also, you mentioned that `the first part always checks for exact match. ex. +331554*, If there is no exact match, it does a regex match`, I don't get that. From the code I can see that if the parameter contains `*` or `#` you try a REGEX search by first removing the `+1`, can you please point me to the part of the code you refer?

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
gvagenas: @croufay what is your comment on that?

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
gvagenas: ok @croufay

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
gvagenas: @croufay can you please explain the steps of your implementation here, I think I am missing the reason and I don't understand it.

gvagenas: @croufay we need a design document here

gvagenas:
@croufay please provide test cases for the REGEX rules we support here. You can copy CallLifeCycleTest or TestDialVerb* and the DB file it uses, to provide REGEX rules in DB instead of full numbers.

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
gvagenas: We should remove the https://github.com/RestComm/Restcomm-Connect/blob/e1f429084ef147ba4b2e1ae78a65e683c037a9e4/restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java#L953-L953 line that adds the +1 to all number that are not formated with internation access code. 
@deruelle please confirm if you agree

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
gvagenas: In order to avoid iterating over ALL IncomingPhoneNumbers we should add a new column at the IncomingPhoneNumbers table to flag REGEX true/false when we create a DID.
There must be a validation method to tell if the number is DID, USSD ShortCode or Regex

gvagenas:
Provide java regex validation at Admin UI and IncomingPhoneNumber REST API

gvagenas:
For the testsuite you should create a new test class similar to CallLifecycleTest, create a new DB based on `restcomm.script_callLifecycleTest` but instead of specific DID numbers, provide the regex patterns.
In the test class provide a simple test case similar to org.restcomm.connect.testsuite.telephony.CallLifecycleTest#testDialClientAlice but every time dial a number that should match the patterns provided in the DB

gvagenas:
@croufay the PR is ready to merge, code is ok, tests are passing, all good.
We are missing documentation, when docs are ready we can merge to master

croufay:
@gvagenas ,

Doc is here, https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api/incoming-phone-numbers-api.adoc#incoming-phone-number-regex-support

gvagenas:
Great, thanks @croufay , I will merge today

PR #1892, opened 2017-02-24
Play video file from RCML

ghjansen:


deruelle:
@hrosa can you review this ?

restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
```diff
@@ -122,7 +122,7 @@
     private final State failed;
 
     // JSR-309 runtime stuff
-    private static final String[] CODEC_POLICY_AUDIO = new String[] { "audio" };
+    private static final String[] CODEC_POLICY_AUDIO = new String[] { "audio", "video" };
```
hrosa: If we always enable audio and video, the declaring codec policy is not needed.
This should probably be configurable.

@ghjansen have you tested all possibly scenarios to understand if having video always enabled may cause problems? Interop issues with PSTN or audio-only systems come to mind. Be careful here.

ghjansen: Thanks @hrosa .

From the behaviour i understood and tested, the `CodecPolicy` acts only as a filter inside JSR 309 driver, allowing or not specific media. In this case, if client's SDP offer contains video information, the driver will allow it and the process continues, playing audio and video. If client's SDP DO NOT contains video information, only the audio from the video file is received by the client. (I tested that with Olympus video and audio call buttons).

And of course, if `CodecPolicy` contains only `audio` then only audio is received no matter the SDP video offer.

I can imagine the configuration that you mentioned as:
1 - include `video` inside `CodecPolicy` only if `restcomm.xml` is configured with `xms`;
2 - create a new configuration inside `restcomm.xml` to allow enable/disable video;
3 - association of option 2 with option 1.

Wdyt?

ghjansen:
Closed without merging via #2453. Related to #2447.

PR #1896, opened 2017-02-27
fix USSD CDR issue not being saved to DB

croufay:
the CDR is now updated in the db for USSD request

### DIFF
--- a/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
+++ b/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
@@ -608,18 +608,8 @@ public class UssdInterpreter extends UntypedActor {
         @SuppressWarnings("unchecked")
         @Override
         public void execute(final Object message) throws Exception {
-            if(logger.isInfoEnabled()) {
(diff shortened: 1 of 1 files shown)


maria-farooq:
@croufay can you please share your findings about root cause why cdr was not being updated.
I see from code that eliminating following 2 conditions you were able to successfully add cdr:

1. `if (CallResponse.class.equals(klass)) {`
2. `if (callInfo.direction().equals("inbound")) {`

eliminating 1 is valid in my opinion as `klass` here can be either one of `CallResponse` or `SipServletRequest`.

But can you explain why you removed second check `if (callInfo.direction().equals("inbound")) `
As per my understanding `inbound` cdr is populated by `Interpreter` and `outbound` cdr is populated by `Call`/`UssdCall`.

But maybe i am missing something so you can help me here..

maria-farooq:

croufay:
@maria-farooq ,

When I sent USSD request, the if (callInfo.direction().equals("inbound")) 

didn't contain the word "inbound" so that part of the code was always false hence, not CDR will be recorded. I think this was because part of the USSD interpreter was copied from the VoiceInterpreter that had the inbound in the callInfo.direction.

Regards


Charles Roufay
Telestax

maria-farooq:
thank for update @croufay i will merge it now

PR #1901, opened 2017-02-28
add ssl support for SMTP and fix issue with TLS over port 587

croufay:
all that needs to be configured is this 
lease note that for TLS the default port is 587 and for SSL it is 465

in the advanced.conf file
SMTP_USER='XXXXXX@gmail.com'
SMTP_PASSWORD='XXXXPASSWORD'
SMTP_HOST='smtp.gmail.com'
SMTP_PORT='587' #If not set default Port will be used. Port 25.


deruelle:
@leftyb can you review and test this on staging ?

leftyb:
Tested and works. Merging.

PR #1924, opened 2017-03-07
Properly handle Akka Actors exceptions - Issue1829

gvagenas:


restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
```diff
@@ -0,0 +1,111 @@
...
+
+    RestcommFaultToleranceStrategy defaultStrategy = new RestcommFaultToleranceStrategy(10, Duration.create("1 minute"),
+            new RestcommFaultToleranceDecider());
+
+    @Override
+    public void onReceive(Object msg) throws Exception {
+        final Class<?> klass = msg.getClass();
+        final ActorRef sender = getSender();
+
+        logger.info(" ********** RestcommSupervisor " + self().path() + " Processing Message: " + klass.getName());
```
maria-farooq: we can make it guarded

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
```diff
@@ -0,0 +1,111 @@
...
+
+        if (msg instanceof Props) {
+            final ActorRef actor = getContext().actorOf((Props) msg);
+            getContext().watch(actor);
+            logger.info("Created and watching actor: " + actor.path().toString());
```
maria-farooq: this log can be at DEBUG level. and we can add it after `liveActors.put(actor.path(), actor);` and then we can add size of liveActors..

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
```diff
@@ -0,0 +1,111 @@
...
+            liveActors.put(actor.path(), actor);
+            sender.tell(actor, getSelf());
+        } else if (msg instanceof Terminated) {
+            logger.info("Received Terminated message");
+            final Terminated t = (Terminated) msg;
```
maria-farooq: this log can be at DEBUG level. and we can add it after `liveActors.remove(t.actor().path());` and then we can add size of liveActors..

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
```diff
@@ -0,0 +1,111 @@
...
+import scala.concurrent.duration.Duration;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+import static akka.actor.SupervisorStrategy.resume;
+
+/**
+ * Created by gvagenas on 22/02/2017.
+ */
+public class RestcommSupervisor extends UntypedActor {
```
maria-farooq: since this class it self seems an actor, what kind of fault-tolerance  will be applied to it?
will it make sense to override postStop method for this actor it self (Assuming this actor should only stop when RC shutdowns, overriding postStop could provide us some stats about liveActors and some other..)

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/faultTolerance/ActorFaultToleranceTest.java
```diff
@@ -0,0 +1,214 @@
...
+import static akka.actor.SupervisorStrategy.stop;
+import static akka.pattern.Patterns.ask;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Created by gvagenas on 22/02/2017.
+ */
+public class ActorFaultToleranceTest {
```
maria-farooq: can we add tests for sending following messages to `RestcommSupervisor` :
- Props (test create actor)
- Terminated (test removal)
- StopChild (test child removal)

maria-farooq:
Hi @gvagenas,
I provided some comments in files.
A part that concerns me a little is following:
```
        ActorRef xyz = null;
        try {
            xyz = (ActorRef) Await.result(ask(supervisor, props, 500), Duration.create(500, TimeUnit.MILLISECONDS));
        } catch (Exception e) {
            logger.error("Problem during creation of actor: "+e);
        }
        return xyz;
```
It make sense to keep this call to synchronous as this is most critical part (to create an actor)
But we have a Single supervisor for whole instance, 
- can you guide a bit that how many simultaneous request it would be able to entertain?
- can we perform loadtest on this branch to check that?
- In a scenario where we get Exception (timeout) is it enough to just print error as on many places we will end up having NPE, (again load test can help here to see how system behaves under pressure)

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
gvagenas: Good point about making the log guarded and adding the sender path

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
gvagenas: Makes sense

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
gvagenas: Makes sense

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
gvagenas: This class will have the default fault-tolerance strategy which is to restart on unhandled exceptions. 
Makes sense to have the postStop to print some info

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/faultTolerance/ActorFaultToleranceTest.java
gvagenas: Props are already tested.
The other two will be added

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
maria-farooq: in that case i think we should handle the exception in `onRecieve` normally with a `try/catch`
as in my opinion we cannot afford to restart this actor, it will cause loosing track of all live actors

gvagenas:
@maria-farooq unfortunately synchronous is the only way to have a custom fault supervisor but we can have several supervisors as we wish. Currently I have provided only one global, since for what we need to do one supervisor is enough, but if later we can provide more.

Also, If we don't use a try/catch there, the calling method will have to handle the exception.

Regarding the performance concerns you mentioned, I plan to run the load tests using this code

gvagenas:
Thanks for the review and the comments @maria-farooq

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
gvagenas: the liveActors HashMap will be removed. It was used only for testing but I have an alternative way to test if an actor restarted or not. No need to keep a HashMap there

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/faulttolerance/RestcommSupervisor.java
gvagenas: Make sense @maria-farooq

PR #1927, opened 2017-03-07
fix NPE issue that prevents USSDPush from working

croufay:
fix NPE issue that prevents USSDPush from working

### DIFF
--- a/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
+++ b/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
@@ -490,7 +490,17 @@ public class UssdInterpreter extends UntypedActor {
                 @SuppressWarnings("unchecked")
                 final CallResponse<CallInfo> response = (CallResponse<CallInfo>) message;
                 // Check from whom is the message (initial call or outbound call) and update info accordingly
-                if (sender == ussdCall) {
+
+                if( response.get().direction().contains("inbound") ){
+                    callInfo = response.get();
+                    ussdCall.tell(new Answer(callInfo.sid()), source);
+                    }
+               if ("outbound-api".equals(response.get().direction())){
+                         outboundCallInfo = response.get();
+                         fsm.transition(message, downloadingRcml);
+                }
+
+                /* if (sender == ussdCall) {
                     callInfo = response.get();
                 } else {
                     outboundCallInfo = response.get();
@@ -502,7 +512,7 @@ public class UssdInterpreter extends UntypedActor {
                 } else {
                     fsm.transition(message, downloadingRcml);
                     //                     fsm.transition(message, initializingCall);
-                }
+                }*/
             }
         } else if (DownloaderResponse.class.equals(klass)) {
             final DownloaderResponse response = (DownloaderResponse) message;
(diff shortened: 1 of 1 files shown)


restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -490,7 +490,17 @@ public void onReceive(final Object message) throws Exception {
                 @SuppressWarnings("unchecked")
                 final CallResponse<CallInfo> response = (CallResponse<CallInfo>) message;
                 // Check from whom is the message (initial call or outbound call) and update info accordingly
-                if (sender == ussdCall) {
+
+                if( response.get().direction().contains("inbound") ){
+                    callInfo = response.get();
+                    ussdCall.tell(new Answer(callInfo.sid()), source);
+                    }
```
maria-farooq: this bracket seems miss indented

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -490,7 +490,17 @@ public void onReceive(final Object message) throws Exception {
...
+               if ("outbound-api".equals(response.get().direction())){
```
maria-farooq: IMO this `if` should be an `else if`.

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -490,7 +490,17 @@ public void onReceive(final Object message) throws Exception {
...
+                         outboundCallInfo = response.get();
+                         fsm.transition(message, downloadingRcml);
+                }
```
maria-farooq: also we can add a last else here in case direction is neither of `outbound-api`, `inbound`.
In that else we can log the direction.
This is for debug purposes to handle all remaining cases.

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -490,7 +490,17 @@ public void onReceive(final Object message) throws Exception {
                 @SuppressWarnings("unchecked")
                 final CallResponse<CallInfo> response = (CallResponse<CallInfo>) message;
                 // Check from whom is the message (initial call or outbound call) and update info accordingly
-                if (sender == ussdCall) {
+
```
maria-farooq: would it make sense to store `response.get().direction()` in a local variable like this:
`final String direction = callInfo.direction();`
and then compare it with possible values.

maria-farooq:
thanks

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
croufay: @maria-farooq 

I am fine with that, please go ahead if you feel that it is a better option.


Charles

maria-farooq: Hey @croufay 
If you agree, please make changes and we can merge this PR
Thanks

PR #1930, opened 2017-03-07
Issue#1928 fix issue that sets From field to null in UssdCallManager.java

croufay:
fix issue that sets From field to null in UssdCallManager.java

### DIFF
--- a/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
+++ b/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
@@ -282,7 +282,7 @@ public class UssdCallManager extends UntypedActor {
         SipURI to = (SipURI)sipFactory.createSipURI(request.to(), uri);
 
         String transport = (to.getTransportParam() != null) ? to.getTransportParam() : "udp";
-        #from = outboundInterface(transport);
+        //from = outboundInterface(transport);
 
         final ActorRef ussdCall = ussdCall();
         final ActorRef self = self();
--- a/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
+++ b/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
@@ -282,7 +282,7 @@ public class UssdCallManager extends UntypedActor {
         SipURI to = (SipURI)sipFactory.createSipURI(request.to(), uri);
 
         String transport = (to.getTransportParam() != null) ? to.getTransportParam() : "udp";
-        from = outboundInterface(transport);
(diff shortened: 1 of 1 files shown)


restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
```diff
@@ -282,7 +282,7 @@ private ActorRef outbound(final Object message) throws ServletParseException {
         SipURI to = (SipURI)sipFactory.createSipURI(request.to(), uri);
 
         String transport = (to.getTransportParam() != null) ? to.getTransportParam() : "udp";
-        from = outboundInterface(transport);
+        //from = outboundInterface(transport);
```
maria-farooq: @croufay if purpose here is to prohibit null assignment in `from` field, then we can do something like this:
```
SipURI obi = outboundInterface(transport);
from = (obi == null) ? from : obi;
```
this way we don't disturb existing logic for execution of `outboundInterface` Method
so if result from `outboundInterface` is null we keep existing value of `from` otherwise we assign `obi` value

croufay: that is fine, I only removed that because the outboundInterface is actually never used. That also works for me.

maria-farooq: @croufay do you mean `outboundInterface ` is not used at all in `USSDCallManger`'s case?
if yes then would it make sense to remove this method? as it is private and this was only reference to this method?

croufay: @maria-farooq that is correct, we could also remove the private method. We already have a check in the API for USSDpush so that the From field cannot be null. 

example:

curl -X POST -H "application/json" http://ACae6e420f425248d6a26948c17a9e2acf:MYPWD@192.168.112-04-24/Accounts/ACae6e420f425248d6a26948c17a9e2acf/UssdPush  -d "To=111111" -d "Url=http://192.168.1.3:8080/restcomm-rvd/services/apps/AP6c234396213e437aa2f8e55c424134ca/controller"
From can not be null.

maria-farooq: @croufay as discussed on Slack:
here is 2 possible solutions:
1. if we want to comment the line like you did then we should remove the unused method `outboundInterface` (as there are no more references to it)
2. if we dont want to disturb any existing use case then we can do what I suggested

I would prefer option 2 as restcomm has its secrets ;)

maria-farooq:
thanks

croufay:
@maria-farooq I have updated the code, please check

PR #1933, opened 2017-03-08
fixed #1456

### LINKED ISSUE #1456: Restrict SupervisorEndpoint and MonitoringEndpoint to SuperAdministrator
These endpoints should be accessible only to top-level accounts i.e. superadministrators.

maria-farooq:
fixed #1456


restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/SupervisorEndpoint.java
```diff
@@ -184,7 +187,7 @@ protected Response registerForUpdates(final String accountSid, final Multivalued
     //Register a remote location where Restcomm will send monitoring updates for a specific Call
     protected Response registerForCallUpdates(final String accountSid, final String callSid, final MultivaluedMap<String, String> data, MediaType responseType) {
         secure(daos.getAccountsDao().getAccount(accountSid), "RestComm:Read:Calls");
```
otsakir: @maria-farooq

Hmm, this line:
> secure(daos.getAccountsDao().getAccount(accountSid), 

 checks the following:

1. User has authenticated
2. Authenticated account is parent of accountSid
3. Authenticated account has the "Restcomm:Read:Calls" permission (as determined by its role)

Then, we also check if this user is a SuperAdmin (i.e. top-level account). 

This means that if a SuperAdmin does not have the "Restcomm:Read:Calls" he won't be granted access. Also, if there is another top-level  account that is not parent (or grant-parent) of the "accountSid" he won't be granted access either. 

Is this ok ? 

Or should we accept that a SuperAdmin has absolute powers on all instance assets and thus rephrase the code as:

// secure()
checkAuthenticatedAccount();
allowOnlySuperAdmin();

I'm not suggesting, i'm just wondering...

maria-farooq: very good point @otsakir I think i would agree that following 2 things are enought to grant access:
1. a valid toke is present
2. it is a super admin.

Since we agree here so letme edit the code

Thanks a lot for your valuable feedback.

maria-farooq: @otsakir code is updated, do you thing we are good to merge here now or we have anything else?
thanks
Maria

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/SupervisorEndpointRoleSensitiveTest.java
```diff
@@ -0,0 +1,113 @@
...
+import com.sun.jersey.api.client.WebResource;
+import com.sun.jersey.api.client.filter.HTTPBasicAuthFilter;
+
+/**
+ * SupervisorEndpointRoleSensitiveTest Supervisor metrices should be accessible to only Super Admin role.
+ *
+ * @author Maria Farooq <dam dam nak nak>
+ */
+@RunWith(Arquillian.class)
+public class SupervisorEndpointRoleSensitiveTest extends EndpointTest {
```
otsakir: I would rename this to SupervisorEndpointAccessControlTest.

otsakir:
@maria-farooq , reviewed. all seems ok to me. I also added another test case for a second top-level account.

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/SupervisorEndpointRoleSensitiveTest.java
maria-farooq: nice!! i like it too letme rename

maria-farooq:
thanks a lot @otsakir for the review. i will merge it now

PR #1990, opened 2017-03-23
issue #1903

leftyb:
Fix SMPP outbound for all languages tested : Greek, japanese, Arabic, chinese.
For INBOUND we need to fix following issue and test again: [issue](https://github.com/RestComm/visual-designer/issues/114)


maria-farooq:
Hi @leftyb can we please check if we can cover this change by a test in test suit and add test for languages we tested. 
reference: `https://github.com/RestComm/Restcomm-Connect/tree/master/restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/sms`

leftyb:
Merging and will add unit test later. So it can be tested and used on staging.

Issue #2178, opened 2017-06-01
RCML Downloader logging does not provide adequate information


otsakir:
When downloader fails to retrieve/parse RCML from a source the exception thrown does not provide enough information in the log to troubleshoot the problem. It's like the following:
```
 [org.restcomm.connect.http.client.Downloader] (RestComm-akka.actor.default-dispatcher-1425) 
Problem while trying to download RCML javax.xml.stream.XMLStreamException: Error parsing the 
RCML:org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Content is not allowed in 
prolog.
```

We could include the following:

1. HTTP return status code
2. URL that was supposed to return proper RCML
3. Small part of the response to get a clue of what went wrong. 50-100 chars only.

deruelle:
We should also probably send a notification back to the developer so he can be notified something is wrong with his application

otsakir:
I'm closing this issue but it's not merged yet. There is another requirement for adding one more notification in case download fails which will be taken care of in another issue, in another sprint.

PR #2213, opened 2017-06-09
Issue2178 better downloader logging

otsakir:
Added more verbose logging when Downloader throws an error. Testsuite has been run locally with no major findings but at the same time comparing results with master branch has not been consistent.

This is about #2178


maria-farooq:
@otsakir just a side comment on RCML download failure.
If download fails with http status notFound then, i see[ VI adds/sends notification](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java#L1811).
otherwise [call fails](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java#L1082).

@gvagenas @deruelle do you think a notification should be added/sent in case of other failures as well (other than NotFound)

maria-farooq:
Quoting @gvagenas [12:22 PM] 

> Oh I see your point. I agree, we should add notification for (2) also.

@otsakir do you think we can sample notification code like we have for case 1 in case 2 as well or you think its too much (in that case we can create a separate issue)

PR #2220, opened 2017-06-10
Document docker usage on OSX

### LINKED ISSUE #2206: Docker Quick Start Guide - Are Forwarded Ports Correct?
I realize there are already a couple of other open issues on this topic ( #10, #44 ) and I am already working on a PR that will hopefully update the docs in a way that would address both of those. 

However, I came across a small issue that I feel may be an error in the existing docs, and I wanted to clarify it before moving forward. 

According to the [Quick Start Guide](
https://github.com/RestComm/Restcomm-Connect/blob/bfb7ab4776d9ca76d86b829f6742991eba449ecc/restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/configuration/docker/Restcomm%20-%20Docker%20Quick%20Start%20Guide.adoc#to-execute-a-command-at-the-container), the command for starting the docker container is: 

`docker run  -i -d --name=restcomm-myInstance -v /var/log/restcomm/:/var/log/restcomm/ -e STATIC_ADDRESS="YOUR_ETHERNET_IP" -e ENVCONFURL="https://raw.githubusercontent.com/RestComm/Restcomm-Docker/master/env_files/restcomm_env_locally.sh" -p 80:80 -p 443:443 -p 9990:9990 -p 5060:5060 -p 5061:5061 -p 5062:5062 -p 5063:5063 -p 5060:5060/udp -p 65000-65050:65000-65050/udp restcomm/restcomm:latest`

I am only interested in the ports here. From what I've seen there's nothing running on port `80` nor port `443`... Rather, as @gvagenas pointed out on the gitter channel, it should be port `8080` that should be forwarded. 

Indeed, when I substituted `-p 80:80 -p 443:443` with `-p 8080:8080` I was able to access restcomm, which makes me think this is just a simple case of a typo, or copy/paste error from somewhere else, rather than anything else... ?

Could someone please confirm this ? Do all three ports need to be forwarded for some other reason? Is there some other env var / config that I'm missing?

( cc'ing @leftyb as well, as @gvagenas pointed out he might be able to help)

Thanks!

gsaslis:
This should address several issues related to running the docker images on OSX. 

Closes #2206, https://github.com/restcomm/Restcomm-Docker/issues/10, https://github.com/restcomm/Restcomm-Docker/issues/44


gsaslis:
@leftyb would you mind taking a look at this please ?

deruelle:
thanks @gsaslis . George, Antonis, Henrique can you also review this ?

deruelle:
@scottbarstow can you check on your side as well ?

scottbarstow:
@deruelle I'll take a look tomorrow

leftyb:
@gsaslis I can not really test if the manual is correct, as I dont use Mac. The other users pointed by Jean can maybe help.

atsakiridis:
@scottbarstow do you plan on checking this? If you guys cannot test I can help since I have a mac, just let me know what you need

gsaslis:
Ok there are a couple of things here: 

## Ports
I am not sure if port `80` and `443` should be included in the docker run command. Even though I have not added a commit on this PR that removes them, afaict, only port `8080` works. 
The secure port is 8443 (instead of 443) but even applying the correct mapping (e.g. `-p 8443:8443`) doesn't work for me. Therefore I removed the comment about secure mode.  
With all this in mind, the `-p 80:80 -p 443:443` should just be replaced with `-p 8080:8080`. 

## `RCBCONF_STATIC_ADDRESS` env var
is this really necessary? I noticed @leftyb didn't include it in his [reply on the issue](https://github.com/RestComm/Restcomm-Connect/issues/2206#issuecomment-307741243)

## Value of `RCBCONF_STATIC_ADDRESS`
I added some comments to help explain what value to pick for this env var on a mac. It'd be worth having someone confirm this. 

## `--net host`
I am starting to understand there is some sort of convention to always run the restcomm docker images with `--net host`? If that is indeed the case, it might be worth completely removing all examples where port forwarding is used, to reduce confusion. I personally prefer the "safer" environment of port forwarding when I run a new "untrusted" container (as would be the case for a "Quick Start Guide" user), but if there's already some convention around this, we might as well externalize it. Offering more options makes onboarding more difficult.

gsaslis:
UPDATE: 
Just came across [this comment](https://github.com/RestComm/Restcomm-Docker/issues/105#issuecomment-290461446) by @scottbarstow explaining we also need to correct the `56xx` ports to `58xx` ports. 

Will add this in a new commit.

deruelle:
@gsaslis thanks for the additional changes to the PR, on the `--net host` it's usually preferred not only because of resources consumption but we can change the ports from the container (like 8080 to 80 and 8443 to 443 and 5080 to 5060...) through configuration so the commands in the documentation wouldn't work anymore if that's the case or that requires extra comments. I was thinking, this would just make it easier for first time users to just go with `--net host` and not worry about port forwarding they might miss. Thoughts ?

gsaslis:
@deruelle agreed. 👍  
Given proper justification (e.g. a small section on "why `--net host`") I think it's a fair enough tradeoff, especially for first time users, as you say.  

In summary, from what I've understood so far, here are the points that should go in that section: 
1. large numbers of ports required for SIP and especially RTP. Meaning it's tedious / not convenient / error prone to include these port mappings all the time, in all your commands / config files / etc. 
(As a side note, [not everyone supports mapping ranges of ports yet](https://github.com/kubernetes/kubernetes/issues/23864) ) 
2. When non-standard ports are used, there's always an extra step (to forward ports from host), again making the process error prone / tedious
3. performance: no need to forward packets one extra step.

I think this type of documentation should be enough for someone who would still prefer the network isolation to go and run this off on his own. For consistency, this should also be made clear throughout the docs (that port forwarding is not the recommended option for RestComm), right?

I guess I now need some feedback on the remaining points.  : )

deruelle:
Thanks @gsaslis !

1. Yeah we were the ones to open that issue :D
2. right
3. Yep. reduces memory usage drastically

> For consistency, this should also be made clear throughout the docs (that port forwarding is not the recommended option for RestComm), right?

Agreed. Would you be so kind to make those changes in the document. Given your new fresh eye as first time user. I think this would make the docs so much better as opposed to us doing the changes as we may overlook too many things

> I guess I now need some feedback on the remaining points. : )

Which ones did I miss ?

gsaslis:
> Would you be so kind to make those changes in the document. 
sure, happy to do so  : ) 

> Which ones did I miss ?
Out of the ones I listed [here](https://github.com/RestComm/Restcomm-Connect/pull/2220#issuecomment-308228193), there's still the ports issue (substituting 80 for 8080 AND removing (?) 443 ) and also the ones about the `RCBCONF_STATIC_ADDRESS` ; )

deruelle:
@gsaslis on ports I would keep the command to 8080 and 8443 which is well know by Java developers and doesn't conflict out of the box with apache if running locally.

on static, @leftyb is correct this is only needed for NATed environments such as AWS so not needed for local install

gsaslis:
In that case, that just leaves the question of why the secure mode (i.e. Port 8443) doesn't work out of the box. 

 Can anybody spot something i'm missing ?

deruelle:
It used to work well on Linux. Can you check if the self certificate generation worked correctly and if the configuration in standalone-sip.xml is correct for the https connector ?

gsaslis:
tbh, i wasn't sure how to check whether the certificate creation worked correctly, so i just looked at the logs: 

```
*** Running /etc/my_init.d/restcommsslconf.sh...
SECURE SELF
Certificate was added to keystore
TRUSTSTORE_FILE is not provided but SECURE is TRUE. We will create and configure self signed certificate
Illegal option:  RSA
keytool -genkeypair [OPTION]...

Generates a key pair

Options:

 -alias <alias>                  alias name of the entry to process
 -keyalg <keyalg>                key algorithm name
 -keysize <keysize>              key bit size
 -sigalg <sigalg>                signature algorithm name
 -destalias <destalias>          destination alias
 -dname <dname>                  distinguished name
 -startdate <startdate>          certificate validity start date/time
 -ext <value>                    X.509 extension
 -validity <valDays>             validity number of days
 -keypass <arg>                  key password
 -keystore <keystore>            keystore name
 -storepass <arg>                keystore password
 -storetype <storetype>          keystore type
 -providername <providername>    provider name
 -providerclass <providerclass>  provider class name
 -providerarg <arg>              provider argument
 -providerpath <pathlist>        provider classpath
 -v                              verbose output
 -protected                      password through protected mechanism

Use "keytool -help" for all available commands

The generated truststore file at /opt/Restcomm-JBoss-AS7/standalone/configuration/restcomm-combined.jks 
```
(not promising  : )  )

also, here's what `/opt/Restcomm-JBoss-AS7/standalone/configuration/standalone-sip.xml` looks like: 

```
...
<!--connector name="https" protocol="HTTP/1.1" scheme="https" socket-binding="https" enable-lookups="false" secure="true">
                <ssl name="https" key-alias="KEY_ALIAS" password="SECRET" certificate-key-file="${jboss.server.config.dir}/keystore.jks" protocol="TLSv1,TLSv1.1,TLSv1.2,SSLv2Hello"/>
            </connector-->
...
```
(commented out for some reason?) 

finally, here's whatever looked relevant from the docker container logs: 

```
20:09:10,976 INFO  [gov.nist.javax.sip.stack.SIPTransactionStack] (MSC service thread 1-2) value -1000 will be used for reliableConnectionKeepAliveTimeout stack property
20:09:10,976 INFO  [gov.nist.javax.sip.stack.SIPTransactionStack] (MSC service thread 1-2) Setting Stack Thread priority to 10
20:09:10,992 WARN  [gov.nist.javax.sip.SipStackImpl] (MSC service thread 1-2) using default tls security policy
20:09:11,010 WARN  [gov.nist.core.net.DefaultSecurityManagerProvider] (MSC service thread 1-2) Using default keystore type jks
20:09:11,011 WARN  [gov.nist.core.net.DefaultSecurityManagerProvider] (MSC service thread 1-2) TLS server settings will be inactive - TLS key store will use JVM defaults keyStoreType=jks javax.net.ssl.keyStore=null javax.net.ssl.keyStorePassword=null
20:09:11,022 INFO  [gov.nist.core.net.DefaultSecurityManagerProvider] (MSC service thread 1-2) javax.net.ssl.trustStorePassword is null, using the password passed through javax.net.ssl.keyStorePassword
20:09:11,024 WARN  [gov.nist.core.net.DefaultSecurityManagerProvider] (MSC service thread 1-2) Using default truststore type jks
20:09:11,025 WARN  [gov.nist.core.net.DefaultSecurityManagerProvider] (MSC service thread 1-2) TLS trust settings will be inactive - TLS trust store will use JVM defaults. trustStoreType=jks javax.net.ssl.trustStore=null javax.net.ssl.trustStorePassword=null
20:09:11,168 INFO  [org.jboss.as.connector.subsystems.datasources] (MSC service thread 1-4) JBAS010400: Bound data source [java:jboss/datasources/ExampleDS]
20:09:11,174 INFO  [org.jboss.as.connector.subsystems.datasources] (MSC service thread 1-3) JBAS010400: Bound data source [java:/MySqlDS]
20:09:11,381 INFO  [org.jboss.as.remoting] (MSC service thread 1-3) JBAS017100: Listening on 172.17.0.3:9999
```

bonus: i found http://documentation.telestax.com/connect/configuration/docker/Restcomm%20-%20Docker%20Using%20Self-signed%20Certificate.html, but that looks a little outdated too (e.g. afaict at least one of the necessary fixes on that page is to move `SECURESSL` to `RCADVCONF_SECURESSL` )?

In any case, I couldn't get the https port to work following the instructions on that page either.

deruelle:
@gsaslis looks like we need to solve on OSX before we can move on further.

> Illegal option:  RSA
> keytool -genkeypair [OPTION]...`

gsaslis:
@deruelle, tbh i'm not sure how this is an osx issue, since that error message is from the docker container logs... : / 

Will try to investigate in more detail tonight / test this on linux too

gsaslis:
UPDATE: 

Fixed `illegal option` message, by fixing var name, in https://github.com/RestComm/Restcomm-Connect/pull/2220/commits/ca3a528e2f117f318635bef04191861667fd4c6e. 

main issue now *seems* to be that the connector is... commented out?

deruelle:
@gsaslis can you please provide a new output copy of the dockers logs in a gist and of the standalone-sip.xml as well ?
If that's the case it may seem that the scripts may need an update to correctly fix the standalone-sip.xml via sed. @leftyb may be able to better comment on this one

gsaslis:
@deruelle sure, here's the gist: 

https://gist.github.com/gsaslis/c3f18dd2bbefdd2646388c6fc0292f7c

deruelle:
@gsaslis it looks like something is wrong still with MacOS for this as on linux the standalone-sip.xml has correctly the value : 

```
<connector name="https" protocol="HTTP/1.1" scheme="https" socket-binding="https" enable-lookups="false" secure="true">
                <ssl name="https" key-alias="restcomm" password="changeme" certificate-key-file="${jboss.server.config.dir}/restcomm-combined.jks" cipher-suite="TLS_RSA_WITH_3DES_EDE_CBC_SHA,TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA256,TLS_RSA_WITH_AES_256_CBC_SHA" verify-client="false" protocol="TLSv1,TLSv1.1,TLSv1.2,SSLv2Hello" />
            </connector>
```

Also in the start logs of Docker, I have something somewhat different see https://gist.github.com/deruelle/2f155613fafef8be22b672d63190ab69

gsaslis:
@deruelle hmmm without going through it in too much detail, i see you're using `restcomm/restcomm:latest`, whereas I was always using the `stable` tag... would you mind checking with that?

deruelle:
Ah sorry @gsaslis don't use stable. See https://hub.docker.com/r/restcomm/restcomm/tags/, please use 8.1.0 or latest.

agafox:
@gsaslis please take a look on this issue if you want to use a :latest https://github.com/RestComm/Restcomm-Docker/pull/109
Also wanted to ask, would you like to add docker-compose.yml with ports and environment variables? It could be a good thing to start Restcomm Docker easily.

gsaslis:
@deruelle wow, then that's another major documentation fail we need to fix here... I immediately assumed `stable` was the safe choice that I should be using (and even [documented this](https://github.com/RestComm/Restcomm-Connect/pull/2220/commits/ae0211a34b2c11f149e4a822c9828e0bc14c01f8) as part of my PR... will need to change that). Imho, we either need to explicitly point out in the docs that it is not, or make the `stable` tag up-to-date. 

Using `latest` does overcome the HTTPS issue I mentioned above. And it also looks like a lot of this could have been avoided... : /

@agafox the docker compose is a good idea. In fact, I would go one step further and we should replace all the spread out `docker run` commands there are lying around with some central doc / bunch of descriptors that explain how to run the container. 

We could then have different options there for: 

- standard `docker run`
- docker compose
- kubernetes
- etc...

deruelle:
@leftyb  @gvagenas any reasons why the `stable` is not matched to the latest releae `8.1.0` in our case ?

on docker compose, let's open a separate issue to brainstorm there and be it a separate PR so we wrap up the current work on what we have today, make it work well on osX as well and have it well documented before improving the image so we iterate on this with clear goals.

deruelle:
@gsaslis we have updated the docker image to fix the WebSockets issue with Olympus. Can you do one last try and see if this pull requests requires more changes to documentation ? Thanks in advance

agafox:
@gsaslis we created new PR with a few fixes of Docker documentation: #2382 
Now, I'm closing your PR because #2382 contains a lot of you changes. But, feel free to create new one if needed. And you also welcome with docker-compose PRs.

gsaslis:
@agafox tbh, i'm not sure why this was closed in favour of #2382 (especially if #2382 contains a lot of the same fixes I originally proposed here)?

wouldn't it make more sense to either: 
a. add extra fixes here (as this also includes all the history), 
b. create another PR with just the different fixes?

Issue #1867, opened 2017-02-21
When using the Search function in Numbers tab, system returns a 404


ipsilantide:
@scottbarstow commented on [Tue Jan 31 2017](https://github.com/RestComm/graphics-design/issues/76)

This is in Search existing numbers, not searching for Provider numbers

Get the following error:

JBWEB000065: HTTP Status 404 - /search

JBWEB000309: type JBWEB000067: Status report

JBWEB000068: message /search

JBWEB000069: description JBWEB000124: The requested resource is not available.

JBoss Web/7.2.0.Final

---

@deruelle commented on [Tue Feb 07 2017](https://github.com/RestComm/graphics-design/issues/76#issuecomment-277938247)

@scottbarstow I can't reproduce on cloud. Can you provide the steps to reproduce and potentially some screenshots ?

---

@scottbarstow commented on [Tue Feb 07 2017](https://github.com/RestComm/graphics-design/issues/76#issuecomment-278009433)

Type a value into search. It filters down. If you then hit enter, you get the error. See attached @deruelle 

![screen shot 2017-02-07 at 9 05 50 am](https://cloud.githubusercontent.com/assets/746403/22694315/bfa3e432-ed14-11e6-94f3-a012fe4ddcc6.png)

![screen shot 2017-02-07 at 9 06 01 am](https://cloud.githubusercontent.com/assets/746403/22694317/c32018ba-ed14-11e6-930a-dd993c6e0a0a.png)

---

@deruelle commented on [Wed Feb 08 2017](https://github.com/RestComm/graphics-design/issues/76#issuecomment-278267927)

Thanks @scottbarstow. Confirmed. Assigning to @brainslog and tied to https://github.com/RestComm/Restcomm-Connect/issues/1455

ipsilantide:
moved here from design repository

muhammadbilal19:
@deruelle this bug is also available in client tab as well. Current functionality is only fetching values from current filter records (which are visible on the current page). On pressing enter button form submit request generated from system which causes 404. 
Do we need to provide remote filtering functionality for Numbers ?

deruelle:
> Do we need to provide remote filtering functionality for Numbers ?

What would be the filtering criterias ?

muhammadbilal19:
@deruelle look like to me only friendly_name and phone_number not sure if customer want any other field for searching in this table.

deruelle:
[...]

PR #2256, opened 2017-06-20
Add filters for IncomingPhoneNumber dashboard ui Issue #1867

muhammadbilal19:
Please have a look at the changes related to remote filtering for IncomingPhoneNumber on dashboard.


deruelle:
Thanks @muhammadbilal19 @ammendonca can you help review this one ?

ammendonca:
Hi @muhammadbilal19 

Thanks for the PR, sorry for the long wait for review.

I'm pasting a screenshot of how it looks after your fix (please try to do this for PRs which change the UI):
image

I think we should keep the same UI/UX, with a single search field that can match either the number or the friendly name, WDYT ? Also, it would be great if the filtering was happening as the user is typing. In summary, keep as it was, with the remote filtering instead of local and ignoring the enter key press.

WDYT ?

cc @ipsilantide

muhammadbilal19:
Thanks @ammendonca sure i will add screenshots next time if i change any UI.

Sounds good to me if we can disable enter key and use only remote filters it looks good to me. As we already implemented pagination and we are not loading all the records on UI so local filtering is useless i guess.
If everyone is happy with this i will update the PR with these changes?

ipsilantide:
+1 to filtering while the user is typing

Issue #1777, opened 2017-02-08
Client Friendly Name is not effected.


maria-farooq:
While creating a new client, an optional parameter is to set friendly name for it:
when we set friendly name it should appear for client in UI. but client login in shown in place of name.
i checked in DB as well and in DB friendly name is set to login, instead of actual value provided while creating client.

```
        friendly_name: client2
                login: client2
```

PR #2266, opened 2017-06-21
Fix Friendly name not updated in case of register new client #1777

muhammadbilal19:
Please have look ui model was mapped with wrong name. Test with Hsql and Mysql both working fine now. Have a look and let me know if you have any comments


deruelle:
Thanks @muhammadbilal19. @ammendonca can you please review this one ?

Issue #2265, opened 2017-06-20
Prevent multiple registration of the same DID 


croufay:
A restcomm client registers 1650825xxxx while another Restcomm client registers +1650825xxxx

The only difference is the plus size but there will be 2 different apps attached to the numbers. This can cause potential problems. The fix is to check if there is a similar number especially if the only difference is the appended + size or the 00 indicating an international number.

result from the DB

select friendly_name, phone_number from restcomm_incoming_phone_numbers where phone_number like '%1650825xxxx%';
+---------------+---------------+
| friendly_name | phone_number  |
+---------------+---------------+
| Charles       | +1650825xxxx  |
| +1650825xxxx  | 1650825xxxx  |
| test          | 001650825xxxx |

marca56:
Need to make sure that this works between and within an organization so that there is only *one* DID in the database overall

deruelle:
@marca56 not sure I understand your comment. Can you elaborate ?

muhammadbilal19:
@croufay @deruelle what i understand in task that we need to restrict creating same number with only difference '+' or '00'
if user already created a number like 1650825xxxx than system will not allow to create a new number like +1650825xxxx or 001650825xxxx and shows error number already exists. Correct me if my understanding is wrong?

deruelle:
@muhammadbilal19 that sounds like a correct statement @muhammadbilal19 do you want to take a crack at it ?

muhammadbilal19:
Yes sure I will work on it

muhammadbilal19:
@croufay @deruelle i created PR for this issue. What about the existing numbers in the DB do we need do something for that as well? or this change is only for new numbers?

deruelle:
@muhammadbilal19 for existing numbers this will need to be done manually as we can't make an arbitrary decision on which number is the valid one I think and we need to warn existing users that have a number conflicting with another one that it will be removed.

muhammadbilal19:
Thanks @deruelle look like that will be handle by RC internal team but still if you need any help from my side let me know once the PR is ready.

PR #2282, opened 2017-06-26
Fix issue of multiple registration of the same DID #2265

muhammadbilal19:
Changes are done for multiple registration issue for incoming numbers. I tried my best to test all possible cases which are know to me. Please have look and let me know if i missed some case.

### DIFF
--- a/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
+++ b/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
@@ -78,7 +78,9 @@ import com.thoughtworks.xstream.XStream;
  * @author quintana.thomas@gmail.com (Thomas Quintana)
  * @author gvagenas@gmail.com
  * @author jean.deruelle@telestax.com
+ * @author muhammadbilal
  */
+
 @NotThreadSafe
 public abstract class IncomingPhoneNumbersEndpoint extends SecuredEndpoint {
     @Context
--- a/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
+++ b/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
@@ -47,6 +47,7 @@ import org.joda.time.DateTime;
 import org.restcomm.connect.commons.annotations.concurrency.NotThreadSafe;
 import org.restcomm.connect.commons.dao.Sid;
 import org.restcomm.connect.commons.loader.ObjectInstantiationException;
+import org.restcomm.connect.commons.util.StringUtils;
 import org.restcomm.connect.dao.DaoManager;
 import org.restcomm.connect.dao.IncomingPhoneNumbersDao;
 import org.restcomm.connect.dao.entities.Account;
@@ -78,7 +79,7 @@ import com.thoughtworks.xstream.XStream;
  * @author quintana.thomas@gmail.com (Thomas Quintana)
  * @author gvagenas@gmail.com
  * @author jean.deruelle@telestax.com
- * @author muhammadbilal
+ * @author muhammad.bilal19@gmail.com (Muhammad Bilal)
  */
 
 @NotThreadSafe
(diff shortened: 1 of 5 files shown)


deruelle:
Thanks @muhammadbilal19. @maria-farooq can you please review this PR ?

maria-farooq:
Hi @muhammadbilal19 
Thanks a lot for your PR

sorry about the delay, we have been under some fires.
Here are a couple of comments that i have after a quick look:

- query change in mysql/mariadb-sql xml file does not seems to match with hsql.
- I see you provided test which are using HSQL (good work) were you able to test mysql query change locally with mysql to verify syntax?
- Did you check, where else in code this query is being used and possible impact or changes required.

Thanks 
Maria

muhammadbilal19:
Thanks @maria-farooq for your feedback. 

- I tested both the queries locally and was working fine for me. Query is change due to syntax and way of working of these two DB's.
- Yes i tested mysql query locally.
- Yes i guess according to my knowledge of the system i checked all the place where this query is used.  

Let me know if you have some specific test case in your mind which i need test.

maria-farooq:
Thanks @muhammadbilal19 

Please check, query for mysql and hsql are logically different.

here is your mysql query
```
 SELECT * FROM restcomm_incoming_phone_numbers WHERE phone_number=#{phone_number}
 	  or  phone_number = "+" + #{phone_number};
```
here is your hsql query
```
    SELECT * FROM "restcomm_incoming_phone_numbers" WHERE "phone_number"=#{phone_number}
   	 or "phone_number" = '+'||#{phone_number}
   	 or "phone_number" = '00'||#{phone_number}
      or "phone_number" = '+'||'00'+#{phone_number};
```
please review again.

Thanks
Maria

muhammadbilal19:
Sorry @maria-farooq for late reply i was bit busy. Yes both queries are different but both works correctly and same. I can understand that you are referring about less WHERE conditions in mysql query but it will return the exact result which we want from that query. 

Say for example i added a number in mysql **006421969195** now if any user want to add this number or any combination of number like **+006421969195** , **6421969195**  and +**6421969195** system will not allowed. 
see below results from mysql query
 
screen shot 2017-07-14 at 10 53 21 pm
screen shot 2017-07-14 at 10 55 43 pm
screen shot 2017-07-14 at 10 57 10 pm
screen shot 2017-07-14 at 10 58 55 pm


This same query was not working for hsql so i need to change it accordingly. So in terms of functionality both are working same and as per my testing according to the functionality which we want to achieve.Hopefully i understand your concern correctly?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
```diff
@@ -327,12 +330,14 @@ protected Response putIncomingPhoneNumber(final String accountSid, final Multiva
...
+        if (isSIP == null) {
             try {
-                number = e164(number);
-            } catch (NumberParseException e) {}
+                number =  e164(number);
+            } catch (NumberParseException e) {
+            }
         }
-        IncomingPhoneNumber incomingPhoneNumber = dao.getIncomingPhoneNumber(number);
+        String phoneFilter  = StringUtils.trimLeadingFrom("+00", number);
```
maria-farooq: @muhammadbilal19 if sequence is `+0` would not it produce same result?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/provisioning/number/vi/IncomingPhoneNumbersEndpointTest.java
```diff
@@ -1004,6 +1007,65 @@ public void testAccountAssociatedPhoneNumbersSecondFilter() {
         assertTrue(clientResponse.getStatus() == 204);
     }
     
+    @Test
+    public void testPurchasePhoneNumberValidation() {
```
maria-farooq: Hi @muhammadbilal19 thanks for adding test case.

Lets say we have following 4 variations of a number:

1. withoutPlusOrZeros (923212345678)
2. withPlus (+923212345678)
3. withDoubleZero (00923212345678)
4. withPlusAndDoubleZero (+00)

Here are possible use cases:

1. Existing number in DB is in format 1, try adding same number with format 1,2,3,4.
2. Existing number in DB is in format 2, try adding same number with format 1,2,3,4.
3. Existing number in DB is in format 3, try adding same number with format 1,2,3,4.
4. Existing number in DB is in format 4, try adding same number with format 1,2,3,4.

Following test case tests use case number 2.
Can you please cover other 3. it would be pretty similar to what you already have.

Please share if you think otherwise.

Regard,
Maria

maria-farooq:
@muhammadbilal19 the queries you changed are corresponding to method `MybatisIncomingPhoneNumbersDao .getIncomingPhoneNumber`. Did you check the references of this method in project to see where-ever it is used, does that part of code need any changes.. for instance CallManager.java

maria-farooq:
Hi @muhammadbilal19  

do we have an update here

Regards
Maria

maria-farooq:
running ci tests on this branch [here](https://mobicents.ci.cloudbees.com/job/RestComm-For-Specific-Branch/86)

maria-farooq:
@muhammadbilal19 following tests are failing on ci for this branch.

maria-farooq:
Hi @muhammadbilal19 

I Hope everything is fine.
Do we have any update here

Maria

PR #2322, opened 2017-07-05
Fixes #1685

### LINKED ISSUE #1685: Make email address for Notification sender configureable.
- https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/BaseVoiceInterpreter.java#L139

hdiniz:
Use existing configurations to set notification sender: smtp-notify/user and smtp-notify/host

### DIFF
--- a/restcomm/restcomm.email/src/main/java/org/restcomm/connect/email/EmailService.java
+++ b/restcomm/restcomm.email/src/main/java/org/restcomm/connect/email/EmailService.java
@@ -175,7 +175,7 @@ public class EmailService extends RestcommUntypedActor {
      EmailResponse send(final Mail mail) {
         try {
             InternetAddress from;
-            if (mail.from() != null || !mail.from().equalsIgnoreCase("")) {
+            if (mail.from() != null && !mail.from().equalsIgnoreCase("")) {
                 from = new InternetAddress(mail.from());
             } else {
                 from = new InternetAddress(user);
--- a/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/BaseVoiceInterpreter.java
+++ b/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/BaseVoiceInterpreter.java
@@ -145,7 +145,7 @@ public abstract class BaseVoiceInterpreter extends RestcommUntypedActor {
     static final int ERROR_NOTIFICATION = 0;
     static final int WARNING_NOTIFICATION = 1;
     static final Pattern PATTERN = Pattern.compile("[\\*#0-9]{1,12}");
-    static String EMAIL_SENDER = "restcomm@restcomm.org";
+    static final String DEFAULT_NOTIFICATION_SENDER = "restcomm@restcomm.org";
 
     // States for the FSM.
     // ==========================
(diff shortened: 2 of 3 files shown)


deruelle:
Thanks @hdiniz ! @leftyb @maria-farooq can you please review as time permits ?

deruelle:
@hdiniz does it requires any changes to restcomm.xml and startup script as well to include the user and host that you used in the code ?

hdiniz:
@deruelle, those settings are already available at [restcomm.xml](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.application/src/main/webapp/WEB-INF/conf/restcomm.xml#L468) and configuration script [config-utensil.sh](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-utensil.sh#L69)

leftyb:
@hdiniz I dont see how the `default-email-address` is configured at the config-utensil.sh.
I think that you need to add a new variable at the email configuration [here](https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/advanced.conf#L63)
And then, as you commented add the necessary modifications at the  config-utensil.sh to set the value at restcomm.xml.

hdiniz:
@leftyb The change was not using `default-email-address`, it was using `smtp-notify/user` and `smtp-notify/host` to form the `From` email header.

Though, now that you pointed that configuration, I agree that it is better. Should I update the change do use `default-email-address` (and add necessary changes for startup)?

leftyb:
@hdiniz I think that using default email address is better yes. 
As well using `smtp-notify/user` and `smtp-notify/host` some times maybe can result to a not valid address. 
Thank you.

hdiniz:
Hi @leftyb, could we go over this again?

Sorry about taking this long, but I was checking this today to update the PR but I found that `smtp-notify/default-email-address` is already used in other context ([RvdProjectsMigrationHelper](https://github.com/RestComm/Restcomm-Connect/blob/55a3f8ab8e111931c4f8c221330c8dc6dc2ebff3/restcomm/restcomm.application/src/main/java/org/restcomm/connect/application/RvdProjectsMigrationHelper.java#L534)). And it is used as the **To Address (destination)** so it could definitely be confusing to use here as the From Address.

In that case as well, `user` and `host` pair from smtp-notify configuration is used to form the From Address.

So I propose either:

1. Keep PR as is, using `user` and `host ` from `smtp-notify` as From Address. **in line with Rvd Migration sender**
2. Add new item to cover this configuration: 
 `smtp-notify/sender-email-address` - **new** - From Address 
 `smtp-notify/default-email-address` - **as is** - To Address

leftyb:
Hi @hdiniz,

Thank you for the PR, I will review and update. 

Regards.

leftyb:
Hi @hdiniz,

I apologize for the delay. I am still bit confused. I guess we want to have a default address (FROM) for the email  if not one is not set no? 

I dont understand when we use the `String notificationSender = DEFAULT_NOTIFICATION_SENDER;`
and when we use the:
 ```
 if (smtpUser != null && !smtpUser.isEmpty()
        && smtpHost != null && !smtpHost.isEmpty()) {
           notificationSender = smtpUser + "@" + smtpHost;
 }
```
It seems that there is a redundancy there. As well What I am saying is that using the HOST & USER is not a good approach. What we want is to make it configurable.  Apart from that using the HOST as explained above can end up with a not valid email address. Host can be anything to e.g. `smtp.mailgun.org`. I would say either use the `DEFAULT_NOTIFICATION_SENDER` but is hardcoded and the issue was all about that. To be able to configure it from the configuration file. 

So I still think that we should add a new configuration as you proposed `sender-email-address - new - From Address` .

Regards.

Issue #2350, opened 2017-07-14
Undeploy/disable RVD if remote deployment provided


otsakir:
If RVD runs remotely (docker), bundled installation should be disabled/undeployed. A This can be handled from a configuration script and create restcomm-rvd.war.undeploy file if RVD_URL parameter is provided in advanced conf.

At the same time, if RVD_URL is not provided the bundler application should be enabled/deployed again.

otsakir:
Issue is fixed.

PR #2351, opened 2017-07-14
Deploy/undeploy bundled RVD through configuration

otsakir:
Added configuration in restcomm-config.sh that controls whether RVD is deployed or not based on the RVD_URL parameter in advanced.conf. By default (RVD_URL is empty) it will get deployed. 

Currently, this PR will make no difference to a a restcomm installation. In an RVD microservice setup when RVD_URL will point to a remote RVD, the bundled version will get undpeloyed.

Refers to #2350


otsakir:
@lefty, i've provided a fix that makes sure RVD is not deployed when restcomm _first_ starts.

PR #2399, opened 2017-07-26
Issue 2389 - API should return 404 if resource is not found instead of 500

maria-farooq:
for update/delete operation of phone sid that does not existed, phone number api was returning 500. updated it to return 404. also added unit tests to verify


maria-farooq:
Adde 2 new tests and verified existing screen shot 2017-07-26 at 1 16 10 pm

PR #2408, opened 2017-08-02
Fix: Go to the next Verb after silent

YevgenL:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -839,7 +840,9 @@ else if (is(gathering) || is(continuousGathering) || (is(finishGathering) && !su
                             }
                         }
                     } else {
-                        collectedDigits.append(data);
```
maria-farooq: Can we add a code comment here for a brief description of this point. so its easier for some future developer to understand. 
Please see other guiding code comment in same method.

PR #2419, opened 2017-08-11
fix issue that creates loop in USSD PUSH

croufay:
I already created a jar that we deployed on the customer server and the problem is resolved. The PR is here for review


abdulazizali77:
@croufay Charles, looks OK to me. I did some minor cleanups to the if block. 
Additionally, can you create a TechDebt issue linked to this to create a regression test for this issue?

PR #2420, opened 2017-08-11
Issue2271 customer work

ghjansen:
This patch contains:
* Fix to continue with next verb after Dial fork receives BUSY from all branches, when RC is using XMS (customer ticket #34390);
* Fix to continue with next verb after Dial fork receives timeout, when RC is using XMS (customer ticket #34415);
* Fix to send statusCallBack notification when dial branch rejects call (customer ticket #34390);
* Fix to consider dial action instead of GetNextVerb, when dial action is configured inside the RCML (found this issue during the optimisation process of previous fixes).


gvagenas:
@ghjansen Please create github issues for each of the problem and provide:
- Clear description of the problem 
- Description of the patch/solution

Also, please prepare unit tests to assert patches. The unit the will be similar to https://github.com/RestComm/Restcomm-Connect/blob/e367a4b3187d37caad0d4d10fa7d906e58e3d02a/restcomm/restcomm.interpreter/src/test/java/org/restcomm/connect/interpreter/GatherSpeechTest.java

ghjansen:
Thank you @gvagenas ! I created the issues #2432, #2433, #2434, #2435 and #2436 following your recommendations. Issue #2436 is in my sprint and i'll start it next week. I'll ping you when ready.

ghjansen:
@gvagenas finally ready. Tests are here: ae97269
I'm not convinced i used the best approach but i put a relevant effort to ensure the 4 main flows behave properly. Please let me know your considerations.

restcomm/restcomm.interpreter/pom.xml
```diff
@@ -172,5 +172,12 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
```
jaimecasero: we should start putting test dependencies in root pom, but this is not particular to this PR

restcomm/restcomm.interpreter/src/test/java/org/restcomm/connect/interpreter/mediagroup/StopMediaGroupTest.java
```diff
@@ -0,0 +1,395 @@
...
+                //this request originally continues through the actors Call > CallController > MediaGroup
+                //reaching media server and stopping whats currently playing. But at this test, both call and
+                //vi actors are using the same observer, so we receive StopMediaGroup directly
+                expectMsgClass(StopMediaGroup.class);
+
+                //emulate the response that comes from media server, through media group, reaching vi
+                interpreter.tell(new MediaGroupResponse<String>("stopped"), observer);
+
+                //check if vi stays still and DO NOT ask for next verb
+                expectNoMsg();
```
jaimecasero: i like this test cases. i find this expectNoMsg a bit opened to changes, or different behaviors, could we add assertions on current state of Actor using "is(State)" method?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -443,6 +457,7 @@ private VoiceInterpreter(VoiceInterpreterParams params) {
         this.asImsUa = params.isAsImsUa();
         this.imsUaLogin = params.getImsUaLogin();
         this.imsUaPassword = params.getImsUaPassword();
+        this.msResponsePending = false;
```
jaimecasero: any reason why we need flags instead of proper state and transitions?. Adding boolean flags to model FSM states will diminish the FSM pattern, and lower maintainability

restcomm/restcomm.interpreter/pom.xml
ghjansen: Agreed. I must admit took this from another PR 😬. I'll move to root then, ok?

restcomm/restcomm.interpreter/src/test/java/org/restcomm/connect/interpreter/mediagroup/StopMediaGroupTest.java
ghjansen: I guess that is not possible given the fact method "is(State)" is not part of the akka actor scope, so no way to invoke it from the test scenario. I'll check alternatives, i agree that checking that also would bring more credibility to the results.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
ghjansen: I tried hard for a pure and dry FSM-based solution. But in the complexity of VI, that was the only functional solution i could came up with. Turns out that all FSM state classes manipulates shared resources that are highly sensitive, so introducing such mechanism was not possible without conflicts that break the call flow. Of course new ideas, alternatives and suggestions are always welcome! (although is important to keep in mind that the current state works!)

jaimecasero: ill have a look,maybe we need to add composite states to the FSM definition...

restcomm/restcomm.interpreter/src/test/java/org/restcomm/connect/interpreter/mediagroup/StopMediaGroupTest.java
jaimecasero: have you tried using TypedActors,instead of untyped?... for test purposes  think is valid and it shoudl only affect on the code that instantiate the actor in the test code, not the actor code itself...

restcomm/restcomm.interpreter/pom.xml
jaimecasero: i did the same for another PR, but im seeing we are all adding test dependencies per subproject, and even worse duplciating the version tag. 
maybe we could add all test dependecies at root level to ensure version consistency, or at least use dependencymanagement to fix version at root level.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
jaimecasero: @ghjansen I can see now that our FSM framework is quite limited (basically not an FSM).

It doesnt even declare how transitions are meant to be fired by FSM depending on incoming events, its just a collection of source and target states. The code needs to specify the transition to take by passing both the event and the target state, which is quite odd for a FSM pattern.

FSM doesnt support composite states. 

The code is also not separating different FSMs properly, so the common use of boolean flags to have different level of states and FSM data model.

Given this FSM status, i would certainly skip/defer any refactoring here, as it will require major effort beyond the scope of this PR.

We should certainly think about refactoring all of this with proper FSM, and potentially evaluate Squirrel as Henrique already did with RMS (all previous features supported...).

restcomm/restcomm.interpreter/pom.xml
ghjansen: Moved to root using `dependencymanagement `.

restcomm/restcomm.interpreter/src/test/java/org/restcomm/connect/interpreter/mediagroup/StopMediaGroupTest.java
ghjansen: I tried to understand how it works @jaimecasero but couldn't make it. I know VI inherits from untyped, but no ideas on how to fit it in the use case avoiding structural changes. (Does it worth the time/effort to only double check FMS state?)

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
ghjansen: Thank you very much @jaimecasero for all your considerations and guidance! Do you think that FSM/Squirrel refactoring would be a good topic to bring to the table in our next R&D offsite? I was thinking about that... not sure.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreterParams.java
```diff
@@ -59,7 +59,7 @@
     private String transferor;
     private String transferee;
 
-    private VoiceInterpreterParams(Configuration configuration, DaoManager storage, ActorRef callManager, ActorRef conferences, ActorRef bridgeManager, ActorRef smsService, Sid account, Sid phone, String version, URI url, String method, URI fallbackUrl, String fallbackMethod, URI statusCallback, String statusCallbackMethod, String referTarget, String emailAddress, ActorRef monitoring, String rcml, boolean asImsUa, String imsUaLogin, String imsUaPassword, String transferor, String transferee) {
+    public VoiceInterpreterParams(Configuration configuration, DaoManager storage, ActorRef callManager, ActorRef conferences, ActorRef bridgeManager, ActorRef smsService, Sid account, Sid phone, String version, URI url, String method, URI fallbackUrl, String fallbackMethod, URI statusCallback, String statusCallbackMethod, String referTarget, String emailAddress, ActorRef monitoring, String rcml, boolean asImsUa, String imsUaLogin, String imsUaPassword, String transferor, String transferee) {
```
jaimecasero: why should we change the constructor modifier? i think there is a bulder available for that,right?

restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
```diff
@@ -496,7 +496,7 @@ private void onCreateMediaSession(CreateMediaSession message, ActorRef self, Act
     }
 
     private void onCloseMediaSession(CloseMediaSession message, ActorRef self, ActorRef sender) throws Exception {
-        if (is(active) || is(initializing) || is(updatingMediaSession)) {
+        if (is(active) || is(initializing) || is(updatingMediaSession) || is(pending)) {
```
jaimecasero: is this "pending"  related to new flag "msResponsePending" ? i was expecting new flag to be here instead of this "pending"... can you clarify relation?

jaimecasero: i see this is different actor, is msResponsePending in voiceInterpreter actor, equivalent to "pending" in jsr309CallController?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
jaimecasero: i've been discussing FSMs with henrique, and specially in the context of RC. He wrote a good summary of RC issues with FSMs two years ago, that is mostly applicable today... 

The custom FSM "framework" we are using today is certainly not meeting the minimum requirements for a proper FSM pattern, just how it enforces you to check the current state, before telling the FSM to transition to target state,instead of just firing the event is painful. 

The great issue though is not the FSM pattern/lib itself, but how we have merged different FSMs into one, that will make any refactoring complex (regardles of which FSM lib we use). The good thing is that we have integration test to partly cover some of these actors behavior. Again, the FSM lib discussion is not key, but how we decouple the monolithic FSMs we have now, into smaller, independent,and cooperative FSMs...

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreterParams.java
ghjansen: Good catch! Residual modification i forgot to clean before commit.

restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
ghjansen: Not exactly. The boolean `msResponsePending` controls the sequence of communication between RC and a MS, while `is(pending)` verification allows the transition from `pending` to `inactive`, properly canceling a call branch. This case is particularly used when a dial fork occurs and a branch that was ringing (pending) answers back with BUSY, so we have to change to inactive. (You can see Henrique's suggestion on this here https://telestax.slack.com/archives/C0D8Y2AAG/p1481801494001255)

Issue #2162, opened 2017-05-18
Ambiguous/misleading CDR status field writing: "COMPLETED" and "completed"


FerUy:
Tied to Zendesk ticket #34316, Restcomm-Connect is generating CDRs with either status **COMPLETED** and **completed**. To maintain coherence, as all other status are written in lower caps,  status field **COMPLETED** should be removed/replaced by **completed**.

FerUy:
@gvagenas I believe the root problem for this issue (which as for the aforementioned Zendesk ticket, it belongs to USSD CDR's), relies in the following two lines of the following XML files (for mariadb & hsql respectively):

https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/call-detail-records.xml#L212

https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.application/src/main/webapp/WEB-INF/sql/call-detail-records.xml#L210

If yes, before making a PR, I would like to understand better the motivation behind `updateInCompleteCallDetailRecordsToCompletedByInstanceId`, as it might make sense for voice calls, but maybe not for USSD.

Thanks in advance!

croufay:
@FerUy ,


Have you tested the above to make sure it works and does what you expect it to do? What happens if you change it to lowercase?



Charles

FerUy:
@croufay, yesterday I had to jump into other issues, so I programmed to test it today... that's why I asked what I asked to @gvagenas, to maybe win some time.

Anyway, after testing, it became clear that indeed this was the issue, and it applies to either voice calls or USSD, so, I believe I should make the PR as for following explanation.

I reproduced the problem with TelScale latest release (8.2.0.1560) and the release generated in my workspace with the aforementioned changes. For example, by generating a USSD session/CDR on "in-progress", then restarting Restcomm-Connect while the call was being processed. While in TelScale latest release (8.2.0.1560) the CDR is rewritten to "COMPLETED", in the release generated in my workspace, it turns to "completed". See images before and after next (with the release generated in my workspace with the aforementioned changes):

*Before* (call being processed, CDR with status **in-progress**):
![before_in-progress](https://user-images.githubusercontent.com/14883482/29142695-18116ece-7d29-11e7-9e34-5d6cd8fdbab0.png)

*After* (CDR with status set to **completed**)::
[...]

Issue #2411, opened 2017-08-03
SIP BYE not answered when sent from USSD Gw after dialogtimeout occurs


FerUy:
A patch for solving https://github.com/RestComm/ussdgateway/issues/69, in order to ending a USSD session which has not been responded by SIP or HTTP to the USSD Gateway accordingly with the `dialogtimeout` value, was tested along with Restcomm-Connect. Same USSD RVD project as the one provided in issue #2410 was used. Likewise, same php script, on this occasion with sleep set to 20 seconds.

**rvd.xml** `externalServiceTimeout` parameter was left to default 5000 value (5 seconds), while `dialogtimeout` in USSD Gw was set to 3000 (3 seconds). Consequently, the expected result was USSD Gw timing out at precisely 3 seconds and therefore sending both MAP End message to SS7 core network and `SIP BYE` to Restcomm-Connect in the meantime the latter was still processing the external service (which was set to answer in 20 seconds, thus Restcomm-Connect would timeout 15 seconds before that, but 2 seconds after USSD Gw's timeout). 

USSD Gw acts as expected, but SIP BYE inbound at Restcomm-Connect doesn't prevent it to keep processing the USSD session and it never answers the multiple SIP BYE messages sent from the USSD Gw after its `dialogtimeout` is reached (3 seconds). Evidence shown in Wireshark trace attached, Restcomm-Connect server.log and RVD RCML log below, which shows the **On Timeout** occurring at 5 seconds, 2 seconds after SIP BYE with correct SIP headers was sent and received from the USSD Gw to end that SIP dialog).

```
[Thu Aug 03 00:14:15 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: start"
[Thu Aug 03 00:14:15 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc][RCML] <Response>
  <UssdCollect action="controller?target=start.step2.handle">
    <UssdMessage>Enter 1, 2 or 3</UssdMessage>
  </UssdCollect>
</Response>
[Thu Aug 03 00:14:19 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: start.step2.handle"
[Thu Aug 03 00:14:19 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: module2"
[...]

Issue #2410, opened 2017-08-02
SIP message is not sent back to USSD Gw when externalServiceTimeout set in rvd.xml is reached


FerUy:
An issue was found when testing **rvd.xml** `externalServiceTimeout` parameter for a USSD Gw <-> Restcomm-Connect implementation for a customer who needs to set a low timeout value for **On timeout** within USSD RVD External Services modules, in order to preventing delayed external services to impact the performance by queuing USSD ongoing sessions.

No matter which value is set, Restcomm-Connect does not proceed with the corresponding SIP message (neither SIP INFO nor SIP BYE depending on the module -*USSD Collect* or *USSD Message* respectively).

The test consisted on a simple USSD RVD project to simply test this (attached). The external service targets a simple php script with a sleep function at the beginning. If the sleep time (1 second) is lower than `externalServiceTimeout` value (2 seconds), the project proceeds to the final USSD Message. This can be easily seen on the USSD RVD project log:

```
[Wed Aug 02 19:58:53 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: start"
[Wed Aug 02 19:58:53 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc][RCML] <Response>
  <UssdCollect action="controller?target=start.step2.handle">
    <UssdMessage>Enter 1, 2 or 3</UssdMessage>
  </UssdCollect>
</Response>
[Wed Aug 02 19:58:56 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: start.step2.handle"
[Wed Aug 02 19:58:56 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: module7"
[Wed Aug 02 19:58:56 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc][ES][REQUEST] "Requesting from url: http://127.0.0.1/sleep.php"
[Wed Aug 02 19:58:57 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc][ES][RESPONSE] "{\"time\":1}"
[Wed Aug 02 19:58:57 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc] "Running target: module5"
[Wed Aug 02 19:58:57 UYT 2017][app APaf6019a57db74ec1bb4bc11cb0fe69bc][RCML] <Response>
[...]

PR #2422, opened 2017-08-14
Change CDR status in uppercase to lowercase for 'completed'

FerUy:
Change CDR status to lowercase **completed** when `updateInCompleteCallDetailRecordsToCompletedByInstanceId` applies. This is for partially solving issue #2162 and related Zendesk ticket #34316. New issues will be raised to solve other problems mentioned in the concerning issue.

### DIFF
--- a/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/call-detail-records.xml
+++ b/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/call-detail-records.xml
@@ -209,7 +209,7 @@
 
   <update id="updateInCompleteCallDetailRecordsToCompletedByInstanceId" parameterType="string">
     UPDATE restcomm_call_detail_records
-      SET status='COMPLETED'
+      SET status='completed'
       WHERE instanceid=#{instanceid} AND (UPPER(status) = ('IN_PROGRESS') OR 
         UPPER(status) = ('IN-PROGRESS') OR UPPER(status) = ('RINGING') OR UPPER(status) = ('QUEUED'));
   </update>
--- a/restcomm/restcomm.application/src/main/webapp/WEB-INF/sql/call-detail-records.xml
+++ b/restcomm/restcomm.application/src/main/webapp/WEB-INF/sql/call-detail-records.xml
@@ -207,7 +207,7 @@
   </update>
   <update id="updateInCompleteCallDetailRecordsToCompletedByInstanceId" parameterType="string">
     UPDATE "restcomm_call_detail_records"
-      SET "status"='COMPLETED'
+      SET "status"='completed'
       WHERE "instanceid"=#{instanceid} AND (UPPER("status") = ('IN_PROGRESS') OR 
         UPPER("status") = ('IN-PROGRESS') OR UPPER("status") = ('RINGING') OR UPPER("status") = ('QUEUED'));
   </update>


abdulazizali77:
@FerUy Fernando, can we in the interim just set it to 'cleanedup' instead of 'completed' ? What do you think?

FerUy:
OK @abdulazizali77, let's wait for @gvagenas input, but I certainly prefer `cleanedup` (or similar) rather than `completed`, as it would put completed and uncompleted transactions in the same category (this is why I created issue #2423 -and related #2424-).

abdulazizali77:
@FerUy Fernando, from a customer point of view/process, are the statistics for calls already exported before any RC nodes are restarted (and hence cleaned up?)

Im trying to understand whether the call statistics before RC restart is important enough to analyze (i assume it should be!) after the restart. Obviously with all the `status`es set to some identical value whether its COMPLETED or CLEANEDUP this would not be possible.

FerUy:
Indeed @abdulazizali77, that's the origin of this (Zendesk ticket #34316). Right now and since the beginning, their IT department consolidates reports everyday and they have a list with a % of USSD sessions `completed` and another % of `COMPLETED` (in their case it's more due to external services in bad shape -answering slowly, generating timeouts, but not creating `canceled` status CDRs as they still don't have RC #2410 and #2411 and https://github.com/RestComm/ussdgateway/issues/69 solved-). 

So, it's important for them to have the proper metrics. Even after these aforementioned issues are solved, they will still eventually get some of these events (much fewer of course), so being able to discriminate a really `completed` USSD session from a `cleanedup` one is important from accounting and statistics perspective (and for us, from a monitoring perspective, as we use CDRs as a monitoring/debugging tool). 

Furthermore, imagine a USSD session that while `in-progress`, an air time or data bundle is attempted to be purchased, and suddenly there is a problem (either in our system or the external service, or due to signaling congestion, or in the radio access). The user will very likely generate a call to customer service (which is expensive for the operator). Then, while in our system the CDR will tell `completed`, in the CRM from the prepaid or postpaid system it will tell otherwise. Got me better now?

abdulazizali77:
@FerUy Thanks Fernando, what you described is exactly the scenario i imagined. But wouldnt it be more visible if we had:
 `in-progress-cleanedup` for call 1 which was `in-progress` and `completed-cleanedup` for call 2 which `completed` 
instead of both `cleanedup` for call 1 and 2?

FerUy:
@abdulazizali77 I'd rather keep it simple, just with `cleanedup` (or similar).

abdulazizali77:
@gvagenas George, Fernando confirms that from this specific customer perspective, setting `cleanedup` to calls that existed before RC restart seems to suffice. 
What are your thoughts? How does this impact RC and other use cases?

gvagenas:
@FerUy @abdulazizali77 to change state `COMPLETED` to `completed` is a minor change and you can proceed.

To introduce a new state `cleanedup`, is a different story and should be carefully designed before you proceed because it will have implications for other services that depend on CDRs, such as Conferencing, thus I believe we shouldn't proceed with the new `cleanedup` state. 

Keep in mind that we are in the process of redesign and implement a new CDR service that will change everything in that space. In that regard, we should invest the minimum time on fixes for the existing CDR service.

abdulazizali77:
thanks @gvagenas @FerUy , merged.

FerUy:
Thanks @gvagenas & @abdulazizali77. For the other situation, you have already created #2423 ;)

PR #2471, opened 2017-09-04
Hints limitation

YevgenL:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
```diff
@@ -124,6 +128,18 @@ private Tag next() {
                             continue;
                         }
                     }
+                    if (tag.name().equals("Gather") && tag.hasAttribute(GatherAttributes.ATTRIBUTE_HINTS) && !StringUtils.isEmpty(tag.attribute(GatherAttributes.ATTRIBUTE_HINTS).value())) {
```
maria-farooq: Please use `Verbs.gather` instead of "Gather" here

Issue #2505, opened 2017-09-19
Tweak config scripts to undeploy RVD for microservice setup


otsakir:
config-restcomm.sh currently relies on RVD_URL variable to determine whether RVD should be undeployed or not. Since relative URLs will be used and RVD_URL variable does not apply any more, the script should rely on new variable RVD_DEPLOY.

PR #2506, opened 2017-09-19
Tweaked configuration scripts to work with RVD_UNDEPLOY env variable

otsakir:
Adds RVD_UNDEPLOY configuration option that controls deploying/undeploying of RVD. Also, it disables RVD configuration in case it is not deployed.

It will be used in microservice setups but also makes sense in setups where RVD is not part of the game.

This relates to issue #2505


otsakir:
Manual testing is currently under way on a new binary.

PR #2508, opened 2017-09-19
wq# This is a combination of 21 commits.

jaimecasero:
two tests refactored for parallel execution

remove Java8 methods

added to maven lifecycle

failsafe plugin moved to plugins,instead of pluginmng

put resource file under original path inside war

more tests refactored

added latest refactored test to failsafe

make tomcatHome unique when forking

make log file name unique across forks

fixed some tests with static ports in urls

add fork id in console output

try in CI with two simple tests

putting debug msg for home

increasing permsize

increased log file size

added tests again

attempt 0

tweaked media port

include all again

fixed xml comment

added some commemts


restcomm/restcomm.application/src/main/java/org/restcomm/connect/application/Bootstrapper.java
```diff
@@ -214,6 +214,7 @@ public UntypedActor create() throws Exception {
 
     private String home(final ServletContext context) {
         final String path = context.getRealPath("/");
+        logger.info("Home path:" + path);
```
maria-farooq: can we use guarded logging here

restcomm/restcomm.testsuite/pom.xml
```diff
@@ -273,18 +273,18 @@
...
-  			<version>1.6.6</version>
-  			<scope>test</scope>
-		</dependency> -->
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.8.5</version>
+            <scope>test</scope>
+        </dependency>	
+        <!-- <dependency>
```
maria-farooq: i believe we can remove the portion of code that is note needed instead of keeping it in comments, i know its coming from master and not this branch, but just a suggestion since its appearing on diff.

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/NetworkPortAssigner.java
```diff
@@ -0,0 +1,111 @@
...
+        File portFile = new File("./target/portFile");
+        try {
+            if (!portFile.exists()) {
+                portFile.createNewFile();
+                LOGGER.info("PID:" + PID + ",portFile created");
+            } else {
+                LOGGER.info("PID:" + PID + ",portFile already exists");
+            }
+        } catch (IOException ex) {
+            LOGGER.info("PID:" + PID + ", there is problem when creating portFile");
```
maria-farooq: maybe we can change it error no?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/ClientsDialAnswerDelayTest.java
```diff
@@ -994,23 +1015,36 @@ public synchronized void testClientWithHostedApplication() throws InterruptedExc
...
-    }
+
+        Map<String,String> replacements = new HashMap();
+        //replace mediaport 2727 
+        replacements.put("2727", String.valueOf(mediaPort)); 
+        replacements.put("8080", String.valueOf(restcommHTTPPort));
+        replacements.put("8090", String.valueOf(mockPort));
+        replacements.put("5080", String.valueOf(restcommPort));
+        replacements.put("5070", String.valueOf(georgePort));        
+        replacements.put("5090", String.valueOf(bobPort));
```
maria-farooq: i dont get why we are replacing client ports, they dont exist in restcomm.xml, or maybe i am missing something

jaimecasero: for some reason the restcom conf uses some of this client ports as outboundProxy....

In addition, we use same this map to replace restcomm conf, db script, and pontentially RCML scripts, so its better to put all the ports there,whether they are goign to effectively change some file, IMO

restcomm/restcomm.application/src/main/java/org/restcomm/connect/application/Bootstrapper.java
jaimecasero: sure,actually it was a logging i put to debug something, i will simply remove it.

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/NetworkPortAssigner.java
jaimecasero: sure

restcomm/restcomm.testsuite/pom.xml
jaimecasero: as you mentioned code is full of commented code, but i can remove this one in particular

PR #2513, opened 2017-09-21
use instance ip as fallback domain name

maria-farooq:


jaimecasero:
for me this PR is ok in terms of providing a fallback uri....

Issue #2477, opened 2017-09-05
Parser should try to validate attribute properties


gvagenas:
like:
- maximum values
- max lenght
- proper URL

etc

carloseluna:
hello, I was reading this topic, but I don't get the request, can you provide me more information about what is required?

carloseluna:
Hello

i was reading the code and I think this is what you need, I only made a validation for "Record", let me know if is this that you need in otder to continue with other Verbs

private Tag next() {
    	boolean shouldContinue=true; // takes the attributes
    	List<Attribute> mapKeys; 

        if (iterator != null) {
            while (iterator.hasNext()) {
                final Tag tag = iterator.next();
                if (Verbs.isVerb(tag)) {
               
                	/********************** add validations to verbs******/
                	if (tag.hasAttributes()) {
                		
                		mapKeys=tag.attributes();
                		
                		if (tag.name().equals("Record")) {
                			
                			for (Attribute e : mapKeys) {
                				 String eName=e.name();
                				 double eValue=Double.parseDouble(e.value());
                				
                				 if(eName.equals("maxLength")) {
                					 if ( (eValue<0) ||(eValue>100) ) {
                						 // this set maxLength between 0 and 100 sec
                						 //if is set tour of those values the process will not go
                						 shouldContinue=false;
                					 }
                				  }
                				 
                				}
                		
                	} 
                }		
               /******************** end validations ***************/	
                	
                    if (current != null && current.hasChildren() && shouldContinue) {
                        final List<Tag> children = current.children();
                        if (children.contains(tag)) {
                            continue;
                        }
                    }
                    current = tag;
                    return current;
                }
            }
        } else {
            if(logger.isInfoEnabled()){
                logger.info("iterator is null");
            }
        }
        return null;
    }

carloseluna:
Hello Guys

I ha been working in this issue, and I attach a file (in txt) with a solution, please review the code and feel free to give me feedback any time

[Perser.txt](https://github.com/RestComm/Restcomm-Connect/files/1320487/Perser.txt)

regards

deruelle:
@carloseluna please make sure to read the [Open Source playbook](https://docs.google.com/document/d/1RZz2nd2ivCK_rg1vKX9ansgNF6NpK_PZl81GxZ2MSnM/edit#heading=h.mcjitemt6ng1) and do a Pull request instead which is the proper procedure for a peer review

deruelle:
[...]

PR #2516, opened 2017-09-25
ISSUE 2477 - Parser should try to validate attribute properties 

carloseluna:
Adding a small solution to issue #2477 making some parsing over the attributes. 

The vertion I had in my repository was a little different from the current, so I adapted my code to the new lines, please feel free to make any comment on it.

### DIFF
--- a/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
+++ b/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
@@ -35,9 +35,14 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Stack;
+import java.util.regex.Matcher; //added
+import java.util.regex.Pattern; //added
+ 
 
 import static javax.xml.stream.XMLStreamConstants.*;
 
@@ -118,11 +123,16 @@ public final class Parser extends RestcommUntypedActor {
     }
 
     private Tag next() throws LimitExceededException{
+		boolean shouldContinue; // new variable
+		
         if (iterator != null) {
             while (iterator.hasNext()) {
                 final Tag tag = iterator.next();
                 if (Verbs.isVerb(tag)) {
-                    if (current != null && current.hasChildren()) {
+					
+					shouldContinue=validation(tag); // add new validation, invoke to class
+					
+                     if (current != null && current.hasChildren() && shouldContinue) {
                         final List<Tag> children = current.children();
                         if (children.contains(tag)) {
                             continue;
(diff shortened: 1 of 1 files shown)


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
```diff
@@ -118,11 +123,16 @@ private void start(final Stack<Tag.Builder> builders, final XMLStreamReader stre
     }
 
     private Tag next() throws LimitExceededException{
+		boolean shouldContinue; // new variable
```
jaimecasero: maybe rename var to "attsValidated" or something that means the attrbiutes were correctly validated

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
```diff
@@ -152,6 +162,235 @@ private Tag next() throws LimitExceededException{
...
+    	boolean shouldContinue=true;
+    	List<Attribute> mapKeys;
+
+    	
+       	if (tag.hasAttributes()) {
+    		
+    		mapKeys=tag.attributes();
+    		
+    		if (tag.name().equals("Record")) {
+    			for (Attribute e : mapKeys) {
```
jaimecasero: This piece of code is rather long, with many if statements (cicloComplexity high), consider refactoring.

can you create an Interface like AttributeValidator, and create as many Validators as required. Then create a Map<String,AttributeValidator> which you can use to find the proper validator instead of so many if statements...

jaimecasero: For every AttributeValidator we need a unitTest exercising the logic, with proper Assertions on when attribute values are right or wrong.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
```diff
@@ -152,6 +162,235 @@ private Tag next() throws LimitExceededException{
...
+    				 String eName=e.name();
+    				 double eValue=Double.parseDouble(e.value());
+    				
+    				 Pattern pattern = Pattern.compile("^([0-9]{1,3})$");
```
jaimecasero: consider reusing the pattern, so we dont need to compile the pattern every time we parse, this is not performance wise...

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
jaimecasero: Since valdiations are of type RegExprPattern, legnth range, and boolean type(this should be regularExpr again), maybe create validators at attribute level, instead of verb level, and create an ConfigurableValidator, which perform the same logic with differetn patterns,length range..., so logic is greatly reduced, and the way the validation error is reported unified...

carloseluna:
Hello guys

I had simplify the code in the parse.java, on the other hand, I had created a Interface class where the system will make all validation, depends on the verb type it is. 

the program will not stop if it see a not match and will return a boolean on parse.java

I reuse a test module you already have in the master, changing some features in my local reposotory. and added code form log4j that you aready have in the original parse. 

Regards

carloseluna:
Hello Guys

I just have added 2 more test cases in order to have a more complete set.

let me know any observation

Regards
Carlos E. Luna

gsaslis:
@jaimecasero could you please take another look here when you get a moment? ;)

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/AttributeValidator.java
```diff
@@ -0,0 +1,548 @@
...
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.log4j.Logger;
+
+public interface AttributeValidator {
+
+	public boolean RecordValidator (Tag tag);
```
jaimecasero: this interface should provide single API for any validator,instead of specific API per RCML verb,that is resolved later using 
[polimorphism](https://stackoverflow.com/questions/1031273/what-is-polymorphism-what-is-it-for-and-how-is-it-used)

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/AttributeValidator.java
```diff
@@ -0,0 +1,548 @@
...
+	public boolean PauseValidator (Tag tag);
+	public boolean SmsAndFaxValidator (Tag tag);
+	public boolean GatherValidator (Tag tag);
+	public boolean PlayAndSayValidator (Tag tag);
+	
+	
+}
+
+ class Validator implements AttributeValidator{
+	   Map <String, String> MyValidators = new TreeMap<String, String>();
```
jaimecasero: this validator class could be a ValidatorBuilder/Factory depending on input tag, back with map poiting to different validator subclasses

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/rcml/Parser.java
```diff
@@ -152,6 +162,47 @@ private Tag next() throws LimitExceededException{
         return null;
     }
 
+	///////////////////////////////////////////// new class -- validate
+    private boolean validation(Tag tag) {
+    	boolean ValidationPass=true;
+    	AttributeValidator AttValidator=new Validator();
+    	
+    	switch (tag.name()) {
```
jaimecasero: i still dont understand why this switch statement is necessary ,and why we need to provide different validator API per RCML verb.I expect a single validator interface, and then different subclasses,invoked with same client code resolved by a map (this is ValidatorFactory responsability).
`
interface AttValidator {
public boolean validate(Tag tag);
}
RecordValidator implements AttValidator{
...
}
ValidatorFactory {
static map<String,AttValidator> validators;
static {
validators.add(new RecordValidator());
...
}
AttValidator retrieveValidator(tag) {
return validators.get(tag.getName());
}
AttValidator validator = validatorFactory.retrieveValidator(tag);
validator,validate(tag);

`

restcomm/restcomm.interpreter/src/test/java/org/restcomm/connect/interpreter/rcml/ParserTest.java
```diff
@@ -106,26 +113,50 @@ public void testParser() {
                 assertTrue(verb.attribute("playBeep").value().equals("false"));
                 parser.tell(next, observer);
                 verb = expectMsgClass(Tag.class);
+                
                 assertTrue(gather.equals(verb.name()));
```
jaimecasero: dont use assertTrue for comparing object use proper assertEquals instead

PR #2527, opened 2017-09-28
Enh parallel4 - testsuite refactored to allow parallel exec

jaimecasero:
latest execution http://ci.telestax.com/view/Restcomm/job/Restcomm-Parallel-Jaime/30/testReport/


jaimecasero:
team decided to merge this asap since only test code is changed

PR #2534, opened 2017-09-29
Issue 2525

maria-farooq:
testsuit is clear for this PR.
@gvagenas @jaimecasero @ghjansen @knosach can anyone please review


restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsEndpointTest.java
```diff
@@ -568,6 +583,79 @@ public void testPasswordStrengthForAccountUpdate() {
         assertEquals(200, response.getStatus());
     }
 
+    @Test
+    public void testCreateAccountInSpecificOrganization() {
+    	// child should not be able to create account in specified org that it does not belong to
```
jaimecasero: this is just a general comment: not sure why you group all this scenarios in one testcase (i think they sequence you are testing now is not related, just in theme), and put a comment. It may be different testcases with proper test intention as part of test case emthod name.

maria-farooq: intention is to test both 
1. permission
2. creation process

reason i would avoid a separate test case is performance, it will take time to gear up a new archive just to separately test above 2 scenarios

restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/accounts.xml
```diff
@@ -28,6 +28,10 @@
 		SELECT * FROM restcomm_accounts WHERE parent_sid=#{parent_sid};
```
jaimecasero: what is this mariadb script? is this something that leaked during testing?

im trying to switch all files generated during test into maven "target" folder, so its inmediately ignore by git, and we dont get into this situations. can we change where the db is started for tests?

maria-farooq: @jaimecasero both mariadb/mysql and hsql have slightly different syntax that is why we have 2 seperate query scripts.
in testsuit we use hsql scripts only

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsEndpointTest.java
jaimecasero: not sure performance is a reason, since i was thinking in creating new test methods/cases inside same test class. For a single test class, there is single container, and deployment of restcomm...but again, this is general comment on testing practices

maria-farooq: right @jaimecasero agreed on practices, i have split them into separate cases

Issue #2503, opened 2017-09-19
PUT method doesn't work in updating a Client's Push flag


atsakiridis:
For example this fails:

```
curl -X PUT https://ACd239d5cab74caaa43a594772e4724687:...@cloud.restcomm.com/restcomm/2012-04-24/Accounts/ACd239d5cab74caaa43a594772e4724687/Clients/CL2735f4b1998c4c6ab8944451503e53ba -d "IsPushEnabled=true"
```

On the other hand POST works fine

PR #2542, opened 2017-10-03
Fixed 'IsPushEnabled' handling in update RC client API call

agafox:


maria-farooq:
would it make sense to add a test case similar to what Antonis mentioned failing in original issue https://github.com/RestComm/Restcomm-Connect/issues/2503

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ClientsEndpoint.java
```diff
@@ -296,8 +297,12 @@ private Client update(final Client client, final MultivaluedMap<String, String>
                 result = result.setVoiceApplicationSid(getSid("VoiceApplicationSid", data));
             }
         }
-        if (data.containsKey("IsPushEnabled") && getBoolean("IsPushEnabled", data)) {
-            result.setPushClientIdentity(client.getSid().toString());
+        if (data.containsKey("IsPushEnabled")) {
+            if (getBoolean("IsPushEnabled", data)) {
+                result = result.setPushClientIdentity(client.getSid().toString());
+            } else {
+                result = result.setPushClientIdentity(null);
```
jaimecasero: i guess this is the only meaningfull change of this PR. This method is strange, it receives a final Client not meant to be modified. Then we assigned to a new var without final, and then we returned the modified instance... not sure what is the reason for this, i know is not part of this PR. Wouldnt make more sense to declare argument as regular non final,rather than suggesting Client is not modified to be modified later...?

jaimecasero:
mostly ok to be merged...

Issue #2444, opened 2017-08-22
UssdInterpreter needs state machine design and implementation cleanup


abdulazizali77:
Some obvious candidates are redundant transition declarations (downloadingRcml ->ready) twice

In relation to https://github.com/RestComm/Restcomm-Connect/pull/2438
Potentially the `sendBye` should be moved to `UssdCall`

PR #2549, opened 2017-10-06
Issue 2521: Move DownloaderResponse access after succeeded() check

abdulazizali77:
@jaimecasero @FerUy Please have a look, apparently the mistake i made before was not moving the `get()` into the `succeeded()` block.
Have added extra logging, advise if overkill.

There is already preexisting TechDebts for `UssdInterpreter` so we will address further possible design issues there separately
https://github.com/RestComm/Restcomm-Connect/issues/2444
https://github.com/RestComm/Restcomm-Connect/issues/2443

### DIFF
--- a/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
+++ b/restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
@@ -545,27 +545,34 @@ public class UssdInterpreter extends RestcommUntypedActor {
             }
         } else if (DownloaderResponse.class.equals(klass)) {
             final DownloaderResponse response = (DownloaderResponse) message;
-
-            int sc = response.get().getStatusCode();
-            //processingInfoRequest
             if (logger.isDebugEnabled()) {
-                logger.debug("Rcml URI : " + response.get().getURI() + "response success=" + response.succeeded()
-                    + ", statusCode=" + response.get().getStatusCode()+" state="+state);
+                logger.debug("Rcml DownloaderResponse success=" + response.succeeded());
             }
             //FIXME: what if these ifblocks arent in downloadingRcml?
             if (response.succeeded()) {
+                int sc = response.get().getStatusCode();
+                //processingInfoRequest
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Rcml DownloaderResponse URI : " + response.get().getURI()
+                        + ", statusCode=" + response.get().getStatusCode()+" state="+state);
+                }
                 if(HttpStatus.SC_OK == sc){
                     fsm.transition(message, ready);
                 } else if (HttpStatus.SC_NOT_FOUND == sc){
                     fsm.transition(message, notFound);
                 }
             } else {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Rcml DownloaderResponse error : " + response.error()
+                        + ", cause=" +
+                        ((response.cause() != null) ? response.cause().getMessage() : ""));
+                }
                 if (downloadingRcml.equals(state) && fallbackUrl!=null) {
                     fsm.transition(message, downloadingFallbackRcml);
                 } else {
                     //unexpected response
                     if(!sentBye && !receivedBye){
-                        sendBye("UssdInterpreter Stopping. Unexpected State when receiving DownloaderResponse");
+                        sendBye("UssdInterpreter Stopping. Unexpected State when receiving DownloaderResponse " + response.error());
                         sentBye = true;
                     }
                     fsm.transition(message, finished);


FerUy:
Looks good to me @abdulazizali77 (and not risky at all), @jaimecasero do you agree?

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -545,27 +545,34 @@ else if ("outbound-api".equals(response.get().direction())){
...
+                }
                 if(HttpStatus.SC_OK == sc){
                     fsm.transition(message, ready);
                 } else if (HttpStatus.SC_NOT_FOUND == sc){
                     fsm.transition(message, notFound);
                 }
             } else {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Rcml DownloaderResponse error : " + response.error()
+                        + ", cause=" + response.cause().getMessage());
```
jaimecasero: im not sure "cuase()" will return an expcetion always in this branch, it may potentially throw NPE...

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -545,27 +545,34 @@ else if ("outbound-api".equals(response.get().direction())){
...
-            int sc = response.get().getStatusCode();
-            //processingInfoRequest
             if (logger.isDebugEnabled()) {
-                logger.debug("Rcml URI : " + response.get().getURI() + "response success=" + response.succeeded()
-                    + ", statusCode=" + response.get().getStatusCode()+" state="+state);
+                logger.debug("Rcml DownloaderResponse success=" + response.succeeded());
             }
             //FIXME: what if these ifblocks arent in downloadingRcml?
             if (response.succeeded()) {
+                int sc = response.get().getStatusCode();
```
jaimecasero: this should be safe now, as i suggested...

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
```diff
@@ -545,27 +545,34 @@ else if ("outbound-api".equals(response.get().direction())){
...
                 }
             } else {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Rcml DownloaderResponse error : " + response.error()
+                        + ", cause=" + response.cause().getMessage());
+                }
                 if (downloadingRcml.equals(state) && fallbackUrl!=null) {
                     fsm.transition(message, downloadingFallbackRcml);
                 } else {
+                    //TODO: should we rethrow response.error()?
```
jaimecasero: exception handling is not always rethrowing... In this case we are both sending BYE potentially,and letting the FSM transition, so i would say no rethrow in this case...

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
abdulazizali77: @jaimecasero 
Seems like theres only two places where `DownloaderResponse` is instantiated, so it doesnt look like anyone every passes a null to exception.
https://github.com/RestComm/Restcomm-Connect/blob/007020dc54fa069ef68ab01424bf17a049cebd9a/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/Downloader.java#L187

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
abdulazizali77: should we remove the TODO?

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
jaimecasero: im talkign abotu this expression "response.cause().getMessage()", and particularly how "response.cause()" may lead to null, not "response" object itself. I just dont want to chagne an NPE for other...

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
jaimecasero: i would certainly remvoe the TODO here

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
abdulazizali77: What i pointed out was there does not seem to be any other Download failure instance where we instantiate the `DownloaderResponse` with null `exception` instead.
Line 187 is the only `DownloaderResponse` failure instance, so when it fails `exception` is always set. Hence there is no failure case where `DownloaderResponse.cause()` is null.

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
abdulazizali77: DONE

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/interpreter/UssdInterpreter.java
abdulazizali77: Made an inline null check, dont really like that idiom for sure, but maybe easiest compromise. Advise if you think we should do away with it.

PR #2564, opened 2017-10-09
Restcomm1147 rms resources leak on conf

gvagenas:
PR also includes MGCP resources metrics and updated tests to assert MGCP resources are properly cleaned up


restcomm/restcomm.testsuite/src/test/resources/log4j.xml
```diff
@@ -1,26 +1,31 @@
...
+		<param name="file" value="target/logs/sip-server${arquillian_sip_port}.log" />
+		<param name="MaxFileSize" value="20000KB" />
+		<!-- Keep one backup file -->
+		<param name="MaxBackupIndex" value="20" />
+		<param name="append" value="false"/>
+		<layout class="org.apache.log4j.PatternLayout">
+			<param name="ConversionPattern" value="%p %t %l - %m%n" />
+		</layout>
+	</appender>
+	<appender name="console" class="org.apache.log4j.ConsoleAppender">
```
jaimecasero: as discussed thsi will create lots of traces in jenkins output...

restcomm/restcomm.testsuite/src/test/resources/log4j.xml.cloud
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
jaimecasero: i guess we need to find a way to configure how this is used in JenkinsCI env...

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
```diff
@@ -89,30 +94,18 @@
...
-    }
-
-    public static Map<MediaSession, ActorRef> getConnections() {
-        return connections;
-    }
-
-    public static Map<MediaSession, ActorRef> getLinks() {
-        return links;
+        connectionToEndpointMap = new ConcurrentHashMap<ActorRef, ActorRef>();
+        connEndpointMap = new ConcurrentHashMap<String, String>();
```
jaimecasero: considering this is Actor state, and concurrency should be already solved by Actor model, is there any particular intention on using ConcurrentHashMap??

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
```diff
@@ -89,30 +94,18 @@
...
-        return endpoints;
-    }
-
-    public static Map<MediaSession, ActorRef> getConnections() {
-        return connections;
-    }
-
-    public static Map<MediaSession, ActorRef> getLinks() {
-        return links;
+        connectionToEndpointMap = new ConcurrentHashMap<ActorRef, ActorRef>();
```
jaimecasero: i dont seem to fine where connectionToEndpointMap  is used? could it be this slid into final code after different design/refactorings?

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
```diff
@@ -400,6 +439,26 @@ private void modifyConnection(final Object message, final ActorRef sender) {
     private void deleteConnection(final Object message, final ActorRef sender) {
         final ActorRef self = self();
         final DeleteConnection dlcx = (DeleteConnection) message;
+        if (dlcx.getConnectionIdentifier() == null) {
+            connEndpointMap.values().removeAll(Collections.singleton(dlcx.getEndpointIdentifier().getLocalEndpointName()));
```
jaimecasero: can we log the result of this operation, i think it could be worth to print it, whether we deleted something or not...

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
```diff
@@ -400,6 +439,26 @@ private void modifyConnection(final Object message, final ActorRef sender) {
...
+            monitoringService.tell(new MgcpConnectionDeleted(null, dlcx.getEndpointIdentifier().getLocalEndpointName()), self());
+            monitoringService.tell(new MgcpEndpointDeleted(dlcx.getEndpointIdentifier().getLocalEndpointName()), self());
+            if (logger.isInfoEnabled()) {
+                String msg = String.format("Endpoint deleted %s", dlcx.getEndpointIdentifier().getLocalEndpointName());
+                logger.info(msg);
+            }
+        } else {
+            connEndpointMap.remove(dlcx.getConnectionIdentifier().toString());
```
jaimecasero: same here, can we save the result of this operation for logging later?

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
```diff
@@ -400,6 +439,26 @@ private void modifyConnection(final Object message, final ActorRef sender) {
...
+            monitoringService.tell(new MgcpConnectionDeleted(dlcx.getConnectionIdentifier().toString(), null), self());
+            if (!connEndpointMap.values().contains(dlcx.getEndpointIdentifier().getLocalEndpointName())) {
```
jaimecasero: could you add some comment to this logic? im not sure im following the intention. It seems we already removed the map entry using the connId, which is the key to the map. Then we check if the endpointId is still in the map as value, if not we tell the service that is was removed. Is there an scenario where connId was removed, but endpointId is still present, what would be the use case there? Are the endpointId shared among different connIds?

restcomm/restcomm.monitoring.service/src/main/java/org/restcomm/connect/monitoringservice/MonitoringService.java
```diff
@@ -113,6 +128,15 @@ public MonitoringService(final DaoManager daoManager) {
         maxConcurrentCalls = new AtomicInteger(0);
         maxConcurrentIncomingCalls = new AtomicInteger(0);
         maxConcurrentOutgoingCalls = new AtomicInteger(0);
+
+        mgcpEndpointsBridge = new AtomicInteger(0);
+        mgcpEndpointsIvr = new AtomicInteger(0);
+        mgcpEndpointsConference = new AtomicInteger(0);
+        mgcpEndpointsPacketRelay = new AtomicInteger(0);
+
+        mgcpEndpointMap = new ConcurrentHashMap<String, String>();
```
jaimecasero: again, do we need concurrent collections as Actor state?

jaimecasero: seems we only need the size of these maps to return as part of onStatistics logic... Do we really need to save all this info into maps? it seems a potential place for unnecesary leaks, when the size of these maps can be already calculated by using other counters already present... I would try to reduce state to simple counters, that maybe be wrong but do not risk to introduce more leaks...

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Call.java
```diff
@@ -1748,17 +1748,8 @@ public void execute(Object message) throws Exception {
...
-                }
-                fsm.transition(message, failed);
-            } else {
-                if (logger.isDebugEnabled()) {
-                    logger.debug("At Call Stopping state, moving to Completed state");
-                }
-                fsm.transition(message, completed);
-            }
+            //Don't wait for ever for the CallController response
+            context().setReceiveTimeout(Duration.create(2000, TimeUnit.SECONDS));
```
jaimecasero: this kind of timeouts should come from configuration, so we can tweak then (or customers) depending on envirnoment... where is 2 secs coming from (for me is regular call timeout, just wondering if this is iline with some other system in the cloud or something)=

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/CallLifecycleTest.java
```diff
@@ -288,6 +289,13 @@ public void testDialCancelBeforeDialingClientAliceAfterTrying() throws ParseExce
         assertTrue(maxConcurrentCalls==0);
         assertTrue(maxConcurrentIncomingCalls==0);
         assertTrue(maxConcurrentOutgoingCalls==0);
+
+        Map<String, Integer> mgcpResources = MonitoringServiceTool.getInstance().getMgcpResources(metrics);
```
jaimecasero: this will be a fairly common construct, could we isolate this in a Assert util class, where access to monitoring services, and literal to specific counters are protected

restcomm/restcomm.testsuite/src/test/resources/log4j.xml.cloud
gvagenas: Agree @jaimecasero . I did this to quickly enable logs in the testsuite since they are essential when working locally with the testsuite

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: @jaimecasero no particular reason, just a habit

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: @jaimecasero I forgot to clean up this one. I will clean this now

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: @jaimecasero you mean the operation to remove from the Map? Yes I can add this at the log statement later

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: Sure as above

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: @jaimecasero RC only deletes connections, on the RMS side when there is no connection for a given endpoint, the endpoint is released.
This is what the method does here.
First remove endpoint-connection, then check if there are more connections for the give endpoint, IF NOT, consider the endpoint released and tell the MonitoringService. 

I added comments for that.

restcomm/restcomm.monitoring.service/src/main/java/org/restcomm/connect/monitoringservice/MonitoringService.java
gvagenas: Its a just a habit

gvagenas: @jaimecasero I am thinking that in the future we can provide MGCP resource details by call, similar to the live call details, so in that case the Map is needed.
Agree with your point but I don't see a risk of leaks here.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Call.java
gvagenas: @jaimecasero this is a safety net for Conference in a cluster configuration where notification from media server might never arrive. It doesn't need to be on configuration since this is an internal, very specific timer to be able to cleanup call no matter what error condition might occur. No particular reason for for the 2 sec duration, I can reduce it if you think its too much

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/CallLifecycleTest.java
gvagenas: @jaimecasero I thought about that but let this task for later as it will take significant time. I did half of the work (`MonitoringServiceTool.getInstance().getMgcpResources(metrics)`) but the rest of the enhancement can be done in another sprint with proper estimation

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Call.java
jaimecasero: 2 seconds seems ok

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/CallLifecycleTest.java
jaimecasero: ok

PR #2611, opened 2017-11-03
new method to count by filter. fixes https://github.com/RestComm/Rest…

jaimecasero:
fixes https://github.com/RestComm/Rescomm-Connect/issues/2610

### DIFF
--- a/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/conference-detail-records.xml
+++ b/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/conference-detail-records.xml
@@ -38,6 +38,33 @@
 		</if>
 		;
 	</select>
+        <!-- use artificial condition (1=1) that always evaluates to true to ease 
+        query composition. 
+        All parameters here are optional.
+        For "status" column use "like" operator to allow finding similar states.
+        -->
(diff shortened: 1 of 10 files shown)


restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/conference-detail-records.xml
```diff
@@ -38,6 +38,33 @@
 		</if>
 		;
 	</select>
+        <!-- use artificial condition (1=1) that always evaluates to true to ease 
+        query composition. 
+        All parameters here are optional.
+        For "status" column use "like" operator to allow finding similar states.
+        -->
+        <select id="countByFilter" parameterType="org.restcomm.connect.dao.entities.ConferenceRecordCountFilter" resultType="int">
+		SELECT COUNT(*) FROM "restcomm_conference_detail_records" WHERE '1' = '1'
```
maria-farooq: can we recheck the syntax for mysql here, i think double quotes are not supported.

jaimecasero: i have just realized too Maria, good catch, i will change the sentence and test it in staging db as you suggested,thanks!!!

PR #2613, opened 2017-11-06
Issue 2590

maria-farooq:


restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh
```diff
@@ -33,25 +33,25 @@ configRestcomm() {
 	static_address="$1"
 
 	sed -e  "s|<\!--.*<external-ip>.*<\/external-ip>.*-->|<external-ip>$static_address<\/external-ip>|" \
-		-e "s|<external-ip>.*<\/external-ip>|<external-ip>$static_address<\/external-ip>|" \
+		-e "s/<external-ip>.*<\/external-ip>/<external-ip>$static_address<\/external-ip>/g;s/<external-ip\/>/<external-ip>$static_address<\/external-ip>/g" \
```
jaimecasero: dont see the point of replacing sed separator from "|" to "/" ,since XML closing tag was already escaped...

jaimecasero: is /g the global substituion flag here? as in https://stackoverflow.com/questions/28231360/command-sed-using-g
what is the purpose of using "/g" here?
are we expecting multiple xml tags with same name in the same line?
If its only to add a new substitution match covering the empty xml tag (<tag/>)?  

would this work as well:
`
sed -e  "s|<\!--.*<external-ip>.*<\/external-ip>.*-->|<external-ip>$static_address<\/external-ip>|" \
		-e "s/<external-ip>.*<\/external-ip>/<external-ip>$static_address<\/external-ip>/" \
                -e "s/<external-ip\/>/<external-ip>$static_address<\/external-ip>/" \
$FILE > $FILE.bak;
`

restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh
```diff
@@ -65,26 +65,36 @@ configRestcomm() {
...
  	fi
 }
 ## Description: OutBoundProxy configuration.
 configOutboundProxy(){
     echo "Configure outbound-proxy"
     FILE=$RESTCOMM_DEPLOY/WEB-INF/conf/restcomm.xml
     sed -e "s|<outbound-proxy-uri>.*<\/outbound-proxy-uri>|<outbound-proxy-uri>$OUTBOUND_PROXY<\/outbound-proxy-uri>|" \
-	-e "s|<outbound-proxy-user>.*<\/outbound-proxy-user>|<outbound-proxy-user>$OUTBOUND_PROXY_USERNAME<\/outbound-proxy-user>|"  \
-	-e "s|<outbound-proxy-password>.*<\/outbound-proxy-password>|<outbound-proxy-password>$OUTBOUND_PROXY_PASSWORD<\/outbound-proxy-password>|" $FILE > $FILE.bak;
+		-e "s|<outbound-proxy-user>.*<\/outbound-proxy-user>|<outbound-proxy-user>$OUTBOUND_PROXY_USERNAME<\/outbound-proxy-user>|"  \
```
jaimecasero: i cant find any change here

jaimecasero:
besides my single comments, i feel even if this PR is working is adding more techDebt than anything.

I didnt manage to find any particular error in sed commands ( as hard as it is to read all commands..., so hard to ensure there are no errors...).

If this PR is just trying to quickly fix this, i'm ok to merge. Otherwise lets try to elaborate on something else (xmlstarlet for all xml edition as we discussed...)

maria-farooq:
@jaimecasero agreed with your analysis, 
sed is very delicate and open to many chances of failure. imagine if a single new character is added restcomm.xml that may fail some part of this script.
not a fan of sed anyway.
a middle term solution can be using xmlstarlet for configuring all xml files (in this case restcomm.xml)
Lets discuss with Marc to see what execution plan makes more sense in short term,
as per long term solution we discussed, most of these scripts will simply disappear anyway.

maria-farooq:
@jaimecasero i have replaced the sed `/` char with `|` as before to lower the risk of having `/` in configured variable, tested locally! will merge it now

PR #2617, opened 2017-11-07
Conference timeout and termination API - Restcomm 1171

maria-farooq:
related issues:
`https://telestax.atlassian.net/browse/RESTCOMM-1171`
`https://telestax.atlassian.net/browse/RESTCOMM-758`


restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/common/http/CustomHttpClientBuilder.java
```diff
@@ -58,12 +68,21 @@ private CustomHttpClientBuilder() {
     }
 
     private static CloseableHttpClient defaultClient = null;
+    private static CloseableHttpAsyncClient closeableHttpAsyncClient = null;
```
jaimecasero: not sure putting sync and async factory in same class is meaningful (maybe a separated async factory would have work better). Lets keep it like this by now...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/asyncclient/HttpAsycClientHelper.java
```diff
@@ -0,0 +1,236 @@
...
+                DownloaderResponse response = new DownloaderResponse(new Exception("got cancellation on executing http request"), "got cancellation on executing http request");
+                sendResponse(sender, response);
+            }};
+    }
+
+    private void sendResponse(final ActorRef sender, Object response){
+        if (sender != null && !sender.isTerminated()) {
+            sender.tell(response, self());
+        } else {
+            if (logger.isInfoEnabled()) {
```
jaimecasero: what happens if we send msg to terminated actor? checking actor state from the outside is not guaranteed,as in the next instruction the state may have changed, so there will be chances where we send the response,and the actor is terminated, not sure is worth to make the distinction here, or maybe send response blindly to actor, and let Akka figure out what to do (deadletter box..)

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/asyncclient/HttpAsycClientHelper.java
```diff
@@ -0,0 +1,236 @@
...
+            try {
+                final Header contentEncoding = entity.getContentEncoding();
+                if (contentEncoding != null) {
+                    builder.setContentEncoding(contentEncoding.getValue());
+                }
+                final Header contentType = entity.getContentType();
+                if (contentType != null) {
+                    builder.setContentType(contentType.getValue());
+                }
+                builder.setContent(StringUtils.toString(stream));
```
jaimecasero: i dont like how response is stringified here, i think it should be original "sender" responsability to do this if appropiate. I can see the existing HttpResponseDescriptor  already expectsa  String content, but i think its a bit dangerous in terms of encoding(this is assuming body is JVM charset encoded/UTF8...) and performance(we loss any streaming chance, and we create an String in memory with potentially big body with no control on actual length)

not sure worth the refactoring, but very worrying...

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
```diff
@@ -577,4 +600,59 @@ private void updateConferenceStatus(ConferenceStateChanged.State state){
...
+     */
+    private long conferenceAge() {
+        if(sid != null){
+            final ConferenceDetailRecordsDao dao = storage.getConferenceDetailRecordsDao();
+            ConferenceDetailRecord cdr = dao.getConferenceDetailRecord(sid);
+            if(cdr == null){
+                if(logger.isDebugEnabled())
+                    logger.debug(String.format("Conference cdr is null for sid: %s", sid));
+            }else{
+                return DateTime.now().getMillis()-cdr.getDateCreated().getMillis();
```
jaimecasero: can we assume all JVM/OS nodes in the cluster are clock synced? are the ndoes configured to sync system clock form a central server regularly?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        assertTrue(georgeCall.waitOutgoingCallResponse(5 * 1000));
+        assertEquals(Response.OK, georgeCall.getLastReceivedResponse().getStatusCode());
+        georgeCall.sendInviteOkAck();
+        assertTrue(!(georgeCall.getLastReceivedResponse().getStatusCode() >= 400));
+
+        String conferenceSid = getConferenceSid(confRoom1);
+        assertEquals(2, getParticipantsSize(conferenceSid));
+        int liveCalls = MonitoringServiceTool.getInstance().getStatistics(deploymentUrl.toString(), adminAccountSid, adminAuthToken);
```
jaimecasero: is it worthit to depend test result on monitorignService? i expect monitoring service to collect on some collection frequency, will the test be stable if we include this assertion here?.

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
```diff
@@ -0,0 +1,350 @@
...
+        if (georgeSipStack != null) {
+            georgeSipStack.dispose();
+        }
+        Thread.sleep(3000);
+        wireMockRule.resetRequests();
+        Thread.sleep(4000);
+    }
+
+    private int getActiveConferencesSize() {
+        JsonObject conferences = RestcommConferenceTool.getInstance().getConferences(deploymentUrl.toString(),adminAccountSid, adminAuthToken);
```
jaimecasero: this may reuse my new countByFilter metthod :)

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
```diff
@@ -0,0 +1,350 @@
...
+        int mgcpConnections = mgcpResources.get("MgcpConnections");
+
+        assertTrue(mgcpEndpoints>0);
+        assertTrue(mgcpConnections>0);
+
+        // Wait for the media to play and the call to hangup.
+        bobCall.listenForDisconnect();
+        georgeCall.listenForDisconnect();
+
+        assertTrue(bobCall.waitForDisconnect(80 * 1000));
```
jaimecasero: i guess this depends on conf timeout, where is the conf timeout configured here? i guess the default 4hours is not good for testing. is this configured at rcml level for this tetscase?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
```diff
@@ -0,0 +1,350 @@
...
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule(8090); // No-args constructor defaults to port 8080
+
+    private static SipStackTool tool1;
+    private static SipStackTool tool3;
+
+    // Bob is a simple SIP Client. Will not register with Restcomm
+    private SipStack bobSipStack;
+    private SipPhone bobPhone;
+    private String bobContact = "sip:bob@127.0.0.1:5090";
```
jaimecasero: can we make this test follow paralle testing approach?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/SecuredEndpoint.java
```diff
@@ -331,6 +331,9 @@ private AuthOutcome secureLevelControl(Account operatedAccount, String resourceA
         String operatedAccountSid = null;
         if (operatedAccount != null)
             operatedAccountSid = operatedAccount.getSid().toString();
+        if (isSuperAdmin()) {
```
jaimecasero: will this affect to other existing operations? Seems like a big change.... Of course this means is the superAdmin anyway, but just wanted to check if this will change anything for other existing RESTAPI operations...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +208,77 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
         }
     }
 
+    protected Response updateConference(String accountSid, String sid, MultivaluedMap<String, String> data,
+            MediaType responseType) {
+        logger.info(String.format("updateConference accountsid: %s conferenceSid: %s", accountSid, sid));
+        Account account = daoManager.getAccountsDao().getAccount(accountSid);
+        try {
+            secure(account, "RestComm:Read:Conferences");
```
jaimecasero: is this the correct literal "RestComm:Read:Conferences"? i dont know the convention, but this is not a ReadOnly operation, here we are modifying the conference, shouldnt be somethign like "Restcomm:Write:Conferences"?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
maria-farooq: @jaimecasero its not 4 hours for testing, its 60 seconds for testing

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: nice catch, updated

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/SecuredEndpoint.java
maria-farooq: this change would mean that superadmin account would have control over restcomm APIs (number, clients, call, conference etc) regardless of the fact that resource being owned by another account in the system

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
maria-farooq: I copied DialConferenceTest as a reference, i will change it but it also means that `DialConferenceTest` (an existing test class in telephony ) is not following parallel approach as well. letme change both

maria-farooq: done

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
maria-farooq: @jaimecasero can you explain how can i use a dao method at this level, is there an api available on top of that ?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
maria-farooq: imo monitoring service is a nice way to check live calls in memory, not sure why test would be unstable.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
maria-farooq: @jaimecasero Network Time Protocol (NTP) is configured by default on Amazon Linux instances and for cloud service is up and running to make it synchronised to NTP server.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/asyncclient/HttpAsycClientHelper.java
maria-farooq: agree with letting akka handle the delivery dynamics

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/asyncclient/HttpAsycClientHelper.java
maria-farooq: should we refactor this as part of this PR?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +207,77 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+            try {
+                secure(account, cdr.getAccountSid(), SecuredType.SECURED_STANDARD);
+            } catch (final AuthorizationException exception) {
+                return status(UNAUTHORIZED).build();
+            }
+
+            final String status = data.getFirst("Status");
+
+            if(status != null){
+                if(status.equalsIgnoreCase("completed")){
```
gvagenas: @maria-farooq I think we need to provide a default response in the case where the provided status DOESN NOT match the `completed`

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +207,77 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+            } else {
+                return null;
+            }
+        }
+    }
+
+    private void kickoutAllActiveParticipants(ConferenceDetailRecord conferenceDetailRecord){
+        List<CallDetailRecord> callDetailRecords = daoManager.getCallDetailRecordsDao().getRunningCallDetailRecordsByConferenceSid(conferenceDetailRecord.getSid());
+
+        if(callDetailRecords == null || callDetailRecords.isEmpty()){
```
gvagenas: @maria-farooq also here I think we need a response to tell user that there were no participants

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +207,77 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+    private ActorRef callApiClient(final Sid sid){
+        final Props props = new Props(new UntypedActorFactory() {
+            private static final long serialVersionUID = 1L;
+
+            @Override
+            public UntypedActor create() throws Exception {
+                return new CallApiClient(sid, daoManager);
+            }
+        });
+        return system.actorOf(props);
```
gvagenas: @maria-farooq here you create a top-level actor with a parent the system itself.

Nowhere in the patch, you destroy this actor, thus this actor and its child actors `HttpAsycClientHelper` will stay in memory for as long as the akka system is up and running.

gvagenas: @maria-farooq I believe better design would be to use ASK pattern, delegate the task to CallManager which create CallApiClient , use to kill the calls and then destroy the actor.

On top, we can use ASK pattern without blocking, just fire the request and return, since we don't need the response.

gvagenas: @maria-farooq also delegating the task to CallManager, we can save some LCM (http) requests.

CallManager can check which of the CDRs are local calls or not and either use update call method or LCM.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    private void onDownloaderResponse(Object message, ActorRef self, ActorRef sender) {
+        final DownloaderResponse response = (DownloaderResponse) message;
+        if (logger.isInfoEnabled()) {
+            logger.info("Call api response succeeded " + response.succeeded());
+            logger.info("Call api response: " + response);
+        }
+        requestee.tell(message, self);
```
gvagenas: @maria-farooq with the current implementation the `requestee` will always be null. 
Shouldn't we protect the code for this?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    protected void onHangup(Hangup message, ActorRef self, ActorRef sender) throws URISyntaxException, ParseException {
+        requestee = sender;
```
gvagenas: @maria-farooq What is the reason behind this? Any plans to use the CallApiClient in other places where we will need to return back the response, because currently there is no use of that

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
```diff
@@ -0,0 +1,163 @@
...
+            public UntypedActor create() throws Exception {
+                return new HttpAsycClientHelper();
+            }
+        });
+        return getContext().actorOf(props);
+
+    }
+
+    @Override
+    public void postStop () {
```
gvagenas: @maria-farooq as far as I can see, CallApiClient will never be destroyed and thus this method will never be called, see my comments on ConferencesEndpoint

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
```diff
@@ -0,0 +1,163 @@
...
+                httpAsycClientHelper = httpAsycClientHelper();
+                HttpRequestDescriptor httpRequestDescriptor =new HttpRequestDescriptor(uri, "POST", postParameters, -1, headers);
+                httpAsycClientHelper.tell(httpRequestDescriptor, self);
+            } catch (Exception e) {
+                logger.error("Exception while trying to terminate call via api: ", e);
+            }
+        }
+    }
+
+    private String getSuperAdminAuthenticationHeader(){
```
gvagenas: @maria-farooq since the ConferencesEndpoint.updateConference() can be used by any account with `RestComm:Modify:Conferences` permision, why we create LCM using super admin credentials? 
I think we should use the credentials of the account that requested the update initially

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
```diff
@@ -0,0 +1,176 @@
...
+                httpAsycClientHelper = httpAsycClientHelper();
+                HttpRequestDescriptor httpRequestDescriptor =new HttpRequestDescriptor(uri, "POST", postParameters, -1, headers);
+                httpAsycClientHelper.tell(httpRequestDescriptor, self);
+            } catch (Exception e) {
+                logger.error("Exception while trying to terminate conference via api: ", e);
+            }
+        }
+    }
+
+    private String getSuperAdminAuthenticationHeader(){
```
gvagenas: @maria-farooq I think we should use the credentials of the account that requested the stop conference initially and not the super admin.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
gvagenas: @maria-farooq I see the local calls are removed by the Conferences endpoint (Evicting private class) but still I think the task to remove remote calls actors should be delegated to CallManager that will create CallApiClient actors and then destroy them

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
```diff
@@ -577,4 +600,59 @@ private void updateConferenceStatus(ConferenceStateChanged.State state){
...
+
+    @Override
+    public void postStop() {
+        if (!fsm.state().equals(uninitialized)) {
+            if(logger.isInfoEnabled()) {
+                logger.info("Conference: " + self().path()
+                    + "At the postStop() method.");
+            }
+            if(conferenceApiClient != null && !conferenceApiClient.isTerminated())
+                getContext().stop(conferenceApiClient);
```
gvagenas: @maria-farooq instead of waiting for the Conference actor to get destroyed in order to destroy the conferenceApiClient, we can make use of the fact the ConferenceApiClient will send the DownloaderResponse to the Confernece actor and kill the actor at this point

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/SecuredEndpoint.java
gvagenas: @maria-farooq but there is already a method for this `org.restcomm.connect.http.SecuredEndpoint#isSuperAdmin` and can be used from any endpoint its needed.

I think if we go with this patch then we mess the SecuredEndpoints and the rules to follow and it will be harder to debug later.

I think we should remove this and instead check with `isSuperAdmin` method

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: done

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/SecuredEndpoint.java
maria-farooq: @gvagenas good point, i added this check for superadmin in conference and call endpoints intead

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: @gvagenas I dont think it need other than a suceesful response, user will get conference resource in response.
this case here is just to make sure in a race case when participants leave themselves before system kick them out

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
jaimecasero: where are those 60 seconds configured?

maria-farooq: in `restcomm-conference.xml` aka restcomm.xml for this specific test class

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/asyncclient/HttpAsycClientHelper.java
jaimecasero: it will impact Downloader as well,so I guess we can do that later...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
jaimecasero: main issue with any single response is that this would require to wait for result on async actions started from here, and which may take time (LVM remote invocation). The response should be succesful if the action of terminating the conf may be initiated.

I think any client can check the RESTAPI later to see how many participants are left at any moment...

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialConferenceTerminationTest.java
jaimecasero: I though monitoringService was getting stats from other components from time to time. Apparently components push data into monitoring service in real time instead. So i guess its safe to use the MonService during test... nevermind my comment

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
maria-farooq: @gvagenas i have changed both `CallApiClient` and `ConferenceApiClient` to be disposable (used for one time request), they will clean them selves after sending back `DownloaderResponse`

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
maria-farooq: i have changed both CallApiClient and ConferenceApiClient to be disposable (used for one time request), they will clean them selves after sending back DownloaderResponse

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
maria-farooq: i have changed both CallApiClient and ConferenceApiClient to be disposable (used for one time request), they will clean them selves after sending back DownloaderResponse

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/CallApiClient.java
maria-farooq: please chek comments below

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
```diff
@@ -309,8 +326,10 @@ public void execute(Object message) throws Exception {
                 final Leave leave = new Leave();
                 call.tell(leave, super.source);
             }
+            //tell conference api client to call conference api to terminate conference and kick all calls
+            conferenceApiClient = conferenceApiClient();
```
gvagenas: @maria-farooq Is there a way to tell if a conference has remote participants or not? This would be a great improvement in order to save resources by not creating and executing the ConferenceApiClient

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
```diff
@@ -1403,9 +1404,45 @@ public void onReceive(final Object message) throws Exception {
...
+        if(message.getCallDetailRecord() == null){
+            if (logger.isDebugEnabled()) {
+                logger.debug("cdr for hangup is null");
+            }
+        } else {
+            // check if it is a local or remote call.
+            if(message.getCallDetailRecord().getInstanceId().equals(RestcommConfiguration.getInstance().getMain().getInstanceId())){
+                ActorRef callTobeHangup = lookup(new GetCall(message.getCallDetailRecord().getCallPath()));
+                if (callTobeHangup != null){
+                    callTobeHangup.tell(message, self);
```
gvagenas: @maria-farooq at the Conference.onStopConference() method, the Conference actor will ask all local call actors to leave.
The call actor at this point will send BYE to the sip client. 
So there is no need to send Hangup to the local call actors.

**Important!** I think there is a bug in the call actor where after Leave message is received, the call actor will never get destroyed.
As far as I can see, when the call actor receives `Leave`message:
1. Will send bye
2. Will tell MmsCallController to leave (destroy media resources)
Possible BUG -> 3. The Call.onMediaServerControllerStateChanged doesn't take care of the MmsCallController response while state is `Leaving`
We need to investigate this one

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
```diff
@@ -553,6 +572,12 @@ private void onStopRecording(StopRecording message, ActorRef self, ActorRef send
         }
     }
 
+    private void onReceiveTimeout(ReceiveTimeout message, ActorRef self, ActorRef sender) throws Exception {
+        if (logger.isInfoEnabled()) {
+            logger.info("Conference Received Timeout, will stop conference now.");
+        }
+        onStopConference(new StopConference(SUPER_ADMIN_ACCOUNT_SID), self, sender);
```
gvagenas: @maria-farooq Maybe we should differentiate the StopConference initiated by SuperAdmin and the one initiated by the timer. This is just an idea, not sure about the implementation but I think if it sounds feasible to you we should investigate it

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private ConferenceDetailRecord getConferenceDetailRecord() throws ParseException{
+        if(conferenceSid == null){
+            ConferenceDetailRecordFilter filter = new ConferenceDetailRecordFilter(accountSid, "RUNNING%", null, null, name, 1, 0);
+            List<ConferenceDetailRecord> conferenceDetailRecords = storage.getConferenceDetailRecordsDao().getConferenceDetailRecords(filter);
+            if(conferenceDetailRecords == null || conferenceDetailRecords.isEmpty()){
+                logger.warning(String.format("could not retrive any conference record for this conference accountSid %s name: %s", accountSid, name));
+                return null;
+            }else{
+                return conferenceDetailRecords.get(0);
```
gvagenas: @maria-farooq isn't this dangerous since we pick the first available Conference in the DB? I believe we should stop and fail the request when conferenceSid is null.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
```diff
@@ -0,0 +1,173 @@
...
+    protected void onStopConference(StopConference message, ActorRef self, ActorRef sender) throws URISyntaxException, ParseException {
+        requestee = sender;
+        conferenceDetailRecord = getConferenceDetailRecord();
+        if(conferenceDetailRecord == null){
+            logger.error("could not retrieve cdr by provided Sid");
+            onDownloaderResponse(new DownloaderResponse(new IllegalArgumentException("could not retrieve cdr by provided Sid")), self, sender);
+        } else {
+            try {
+                URI uri = new URI("/restcomm"+conferenceDetailRecord.getUri());
+                uri = UriUtils.resolve(uri);
```
gvagenas: @maria-farooq here we rely on the fact that the cluster has a common public IP address (the LB public IP). Nothing bad in the current phase but maybe the `conferenceDetailRecord.getUri()` should contain the remote IP Address (when/if the conference is remote).
This would help later to have conferences between instances that don't belong to the same "cluster". 
This is just an idea for discussion.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +202,75 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
         }
     }
 
+    protected Response updateConference(String accountSid, String sid, MultivaluedMap<String, String> data,
+            MediaType responseType) {
+        logger.info(String.format("updateConference accountsid: %s conferenceSid: %s", accountSid, sid));
+        Account account = daoManager.getAccountsDao().getAccount(accountSid);
+        Account effectiveAccount = userIdentityContext.getEffectiveAccount();
+        try {
+            secure(account, "RestComm:Modify:Conferences");
```
gvagenas: @maria-farooq Since later we use `isSuperAdmin()` I don't think we need to check permissions here.
The `updateConference()` method could be protected with just the `isSuperAdmin()`

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +202,75 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+        }
+        final ConferenceDetailRecordsDao dao = daoManager.getConferenceDetailRecordsDao();
+        ConferenceDetailRecord cdr = dao.getConferenceDetailRecord(new Sid(sid));
+        if (cdr == null) {
+            return status(NOT_FOUND).build();
+        } else {
+            //allow super admin to perform update on any conference - https://telestax.atlassian.net/browse/RESTCOMM-1171
+            if (!isSuperAdmin()) {
+                try {
+                    secure(account, cdr.getAccountSid(), SecuredType.SECURED_STANDARD);
```
gvagenas: @maria-farooq can you elaborate on that? Why we use secure again here?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +202,75 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+                } catch (final AuthorizationException exception) {
+                    return status(UNAUTHORIZED).build();
+                }
+            }
+
+            final String status = data.getFirst("Status");
+
+            if(status != null){
+                if (status.equalsIgnoreCase("completed")) {
```
gvagenas: @maria-farooq you can combine `status != null` AND `status.equalsIgnoreCase("completed")` in the same `if` statement so we don't need duplicate `else` clause that does exactly the same thing

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +202,75 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+                logger.debug("no active participants found.");
+        } else {
+           try {
+               if (logger.isDebugEnabled())
+                    logger.debug("total conference participants are: "+callDetailRecords.size());
+               Iterator<CallDetailRecord> iterator = callDetailRecords.iterator();
+                while(iterator.hasNext()){
+                    final CallDetailRecord CallDR = iterator.next();
+                    if (callManager == null)
+                        callManager = (ActorRef) context.getAttribute("org.restcomm.connect.telephony.CallManager");
```
gvagenas: @maria-farooq why don't you retrieve the `CallManager` actor outside the `while` statement so you do it just once and not for every CDR in the list?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
```diff
@@ -194,4 +202,75 @@ protected Response getConferences(final String accountSid, UriInfo info, MediaTy
...
+                    callManager.tell(new Hangup("Conference Terminated", effectiveAccount.getSid(), CallDR), null);
```
gvagenas: @maria-farooq Does it worth to create a new message instead of use Hangup that is used by VoiceInterpreter to ask the Call actor to terminate the call? 
Not important, just wondering if it would be a better approach to have a new message here

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
maria-farooq: we can provide a `reason` field in `StopConference` as well in API,  in case of timeout its will be `conference time limit reached` this field will later be passed as reason header in corresponding BYE to sip clients

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
maria-farooq: we can check by DB but that is like alternative to me not an optimisation, in most cases it would be an overhead query

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
maria-farooq: `conferenceSid ` is not always available at local conference we care about `accountSid:name`

gvagenas: @maria-farooq but still picking up the first on the list its dangerous since this one might not be the conference we are looking for

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: good point

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: good point, changed

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: conference termination request can be initiated by owner of the conference as well hence we need to make sure that resource is being accessed by proper account

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ConferencesEndpoint.java
maria-farooq: conference termination request can be initiated by owner of the conference as well hence we need to make sure that resource is being accessed by proper account

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
maria-farooq: @gvagenas as per our conference design/state, there is only one conference in DB in running state.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
maria-farooq: @gvagenas do you think this is part of the scop of this PR? this bug is there regardless if we merge this patch or not no? can we handle that seperatly

gvagenas: @maria-farooq `do you think this is part of the scop of this PR?` the bug is not in the scope of this PR but I believe CallManager shouldn't try to Hangup local calls.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
gvagenas: @maria-farooq ??? Only one conference in running state? I am missing something here.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
gvagenas: @maria-farooq sounds good

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/api/ConferenceApiClient.java
gvagenas: There will be only 1 RUNNING Conference for AccountSid and given Name
Conference ID is unique because AccountSid+ConferenceName is unique

gvagenas:
@maria-farooq I am not comfortable with the current implementation because I can see redundant functionality and not a clear separation of concerns in the actors/endpoints involved. Let me elaborate:

- The feature here is about how to terminate a Conference and remove all call actors.
- The conference termination might be initiated because of a timer or because of a user action (REST API)
- Call actors might be either local actors or remote actors.
- Local actors should be removed using Akka messaging (send `Leave` message)
- Remote actors should be removed using LCM

**Currently,** your patch implements the following:
1. On timer timeout `Conference` actor moves to `Evicting` state where a) asks Local actors to `Leave` b) creates a `ConferenceApiClient` actor and send to it the `StopConference` message
2. The `ConferenceApiClient` will then create a REST API request for the `ConferencesEndpoint` to update Conference.
3. The `ConferencesEndpoint` will ask `CallManager` to `Hangup` all calls found in the CDR table related to the given Conference
4. `CallManager` will first try to remove local calls and then will try to remove local calls using `CallsApiClient`
5. `CallApiClient` will create an LCM REST API request to `CallsEndpoint` to complete the remote call

**My comments on the above implementation:**
1. When `Conference` actor reach the `Evicting` state, should be self-sufficient and a) remove local call actor b) check DB and remove remote call actors using LCM
1. In both timeout and user-initiated StopConfernece, Conference actor should be responsible to move FSM state to `Evicting` and thus remove local and remote call actors
1. The `ConferenceApiClient` is redundant in that case, it shouldn't be used
1. On user-initiated `StopConference` the `ConferencesEndpoint` should not try to kick out any participant, this endpoint role is not to deal with participants and calls. The `ConferencesEndpoint` should only try to update the state of the Conference to `Evicting` by sending `StopConference` to either the local or remote Conference actor (akka messaging or LCM)
1. This way, we don't have duplicate functionality where ConferencesEndpoint kicks out participants, its the responsibility of Conference actor to kick out call actors
1. Thus `CallApiClient` is also redundant here
1. LCM needs to be enhanced in order to: 
  1. When we complete a call actor because Conference stopped, we need the proper Header in the Bye request
  2. Be able to reach remote Conference actors and send to them `StopConference`

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
```diff
@@ -95,6 +108,8 @@
     private Sid sid;
     private final List<ActorRef> calls;
     private final List<ActorRef> observers;
+    //list of callApiClients created by Conference
+    private List<ActorRef> callApiClients;
```
gvagenas: @maria-farooq There is no need to keep a list of CallApiClients, this is redundant and will consume memory for no reason. See my comment later for the `onDownloaderResponse` method

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Conference.java
```diff
@@ -553,6 +581,26 @@ private void onStopRecording(StopRecording message, ActorRef self, ActorRef send
...
+        if (logger.isInfoEnabled()) {
+            logger.info("Conference Received Timeout, will stop conference now.");
+        }
+        onStopConference(new StopConference(SUPER_ADMIN_ACCOUNT_SID), self, sender);
+    }
+
+    private void onDownloaderResponse(DownloaderResponse message, ActorRef self, ActorRef sender) {
+        //make sure we only remove sender that is callApiClient and was created by this conference actor
+        //so we don't kill a sender that has accidently broadcast a DownloaderResponse to this conference.
+        if(callApiClients.contains(sender)){
```
gvagenas: @maria-farooq as said earlier no need to keep a list of `CallApiClient` and then check if sender is on the list.
I believe we should stop sender straight away UNLESS if there is a chance to receive DownloaderResponse from other actors, which I cannot think of such a case.

Even if that's the case (other actors sends DownloaderResponse to Conference) maybe its better to create another wrapper message for the responses of `CallApiClient`

maria-farooq:
@gvagenas created a new wrapper for callapiresponse instead of reusing downloader response.

PR #2619, opened 2017-11-07
Record race condition - Restcomm-1219

gvagenas:
Patch to fix Record race condition.


restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
```diff
@@ -110,55 +117,62 @@
     private ActorRef monitoringService;
 
     private File recordingFile = null;
+    private int recordingMaxLength;
+
+    private DateTime startTime;
+    private DateTime endTime;
 
     private ActorSystem system;
+    private NotificationRequest recordingRqnt;
```
jaimecasero: maybe we can refactor this on next iteration to have a map of ongoing recordingRqnt and associated sender with reqId as key, so we may notify later on any scenario (including multiple RQNY by multiple RQNT senders) ...

Currently this seems to specific to record scenario...

restcomm/restcomm.mscontrol.mms/src/main/java/org/restcomm/connect/mscontrol/mms/MgcpMediaGroup.java
```diff
@@ -253,8 +255,12 @@ protected void notification(final Object message) {
         } else {
             event = new MediaGroupResponse<>(response.cause(), response.error());
         }
-        if (originator != null)
+        if (originator != null) {
             this.originator.tell(event, self);
+            if (recording) {
+                recording = false;
```
jaimecasero: set to false only if originator is not null? is this right? shouldnt we set to false in any case since we have the ivrresponse?

restcomm/restcomm.mscontrol.mms/src/main/java/org/restcomm/connect/mscontrol/mms/MmsBridgeController.java
```diff
@@ -247,6 +254,19 @@ public void onReceive(Object message) throws Exception {
...
+            onMediaGroupResponse((MediaGroupResponse)message);
+        }
+    }
+
+    private void onMediaGroupResponse(MediaGroupResponse message) throws TransitionNotFoundException, TransitionFailedException, TransitionRollbackException {
+        if (logger.isInfoEnabled()) {
+            logger.info("Received MgcpGroupResponse: "+message.toString());
+        }
+        if (recording && stoppingStatePending) {
+            recording = Boolean.FALSE;
```
jaimecasero: shoudlnt this flags be set with appriate single condition,instead of composite condition like:
`
if (recording && stoppingStatePending) fsm.trastion(message,stopping);
if (recording) recording =false;
if(stoppingStatePending) stoppingStatePending = false;
`

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: @jaimecasero yes that could be needed for parallel testing indeed. I can do it now so we can enable parallel testing for this test class

restcomm/restcomm.mscontrol.mms/src/main/java/org/restcomm/connect/mscontrol/mms/MgcpMediaGroup.java
```diff
@@ -123,6 +123,8 @@
     protected final String primaryEndpointId;
     protected final String secondaryEndpointId;
 
+    private boolean recording;
```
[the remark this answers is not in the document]
gvagenas: @jaimecasero agree

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/MockMediaGateway.java
gvagenas: @jaimecasero actually parallel testing for this class wont be able because we rely on `ReceiveTimeout` to mock the maxLength attribute of Recording. 
In case of parallel testing, the `ReceiveTimeout` timer will never hit.

I live this for next iteration

PR #2635, opened 2017-11-15
Issue1233 Customer-created regex in their account of “.” was swallowing all inbound traffic (at least for SMS)

jaimecasero:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private IncomingPhoneNumber findSingleNumber(String number,
+            Sid sourceOrganizationSid, Sid destinationOrganizationSid) {
+        IncomingPhoneNumber matchedNumber = null;
+        IncomingPhoneNumberFilter.Builder filterBuilder = IncomingPhoneNumberFilter.Builder.builder();
+        filterBuilder.byPhoneNumber(number);
+        if (destinationOrganizationSid != null) {
+            filterBuilder.byOrgSid(destinationOrganizationSid.toString());
+        }
+        if (sourceOrganizationSid!= null &&
```
gvagenas: @jaimecasero can you elaborate on the reason behind this rule?
Is it?
Client XYZ the belongs to Org A dial to DID +123456789 that belongs to Org B? and thus since we don't allow calling sip numbers between Organizations the number to be matched must NOT BE pureSip ?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
```diff
@@ -0,0 +1,245 @@
...
+            logger.debug("getMostOptimalIncomingPhoneNumber: " + phone
+                    + ",srcOrg:" + sourceOrganizationSid
+                    + ",destOrg:" + destinationOrganizationSid);
+        }
+        List<String> numberQueries = createPhoneQuery(phone);
+
+        IncomingPhoneNumber numberfound = findByNumber(numberQueries, sourceOrganizationSid, destinationOrganizationSid);
+        if (numberfound == null) {
+            //if no perfect match found, try with special star number
+            numberfound = findSingleNumber("*", sourceOrganizationSid, destinationOrganizationSid);
```
gvagenas: @jaimecasero I think REGEX numbers should have higher priority than the `*` which should be the last choice if nothing else matched.
The order should be from the most specific to the catch-em-all option, so:
specific number -> regex -> '*'

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
```diff
@@ -0,0 +1,245 @@
...
+     * @param phone
+     * @param numbersDao
+     * @return
+     */
+    private IncomingPhoneNumber findByRegex(List<String> numberQueries,
+            Sid sourceOrganizationSid, Sid destOrg) {
+        IncomingPhoneNumber numberFound = null;
+        IncomingPhoneNumberFilter.Builder filterBuilder = IncomingPhoneNumberFilter.Builder.builder();
+        filterBuilder.byOrgSid(destOrg.toString());
+        filterBuilder.byPureSIP(Boolean.FALSE);
```
gvagenas: @jaimecasero why we exclude pure sip numbers here? 
The Regex numbers will be created as pure sip numbers since they will be created manually and not through a Voip provider integration

restcomm/restcomm.telephony/resources/mybatis.xml
```diff
@@ -19,26 +19,28 @@
...
-    <mapper resource="mappers/sand-boxes.xml"/>
-    <mapper resource="mappers/short-codes.xml"/>
-    <mapper resource="mappers/sms-messages.xml"/>
-    <mapper resource="mappers/transcriptions.xml"/>
-    <mapper resource="mappers/gateways.xml"/>
-    <mapper resource="mappers/announcements.xml"/>
-    <mapper resource="mappers/extensions-configuration.xml"/>
-    <mapper resource="mappers/media-servers.xml"/>
-    <mapper resource="mappers/media-resource-broker-entity.xml"/>
+    <mapper resource="org/restcomm/connect/dao/mybatis/accounts.xml"/>
```
gvagenas: @jaimecasero why you point to the dao resources mappers? Those are also copies of the proper mappers that are located to:
restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/sql
and
restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql

I believe when you apply the fix for HSQL to have common mappers, mappers should be moveds to the resources of the restcomm.application and all tests should point to them as you said back then

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/mybatis/MybatisIncomingPhoneNumbersDao.java
```diff
@@ -326,7 +234,7 @@ private IncomingPhoneNumber toIncomingPhoneNumber(final Map<String, Object> map)
         map.put("date_updated", DaoUtils.writeDateTime(incomingPhoneNumber.getDateUpdated()));
         map.put("friendly_name", incomingPhoneNumber.getFriendlyName());
         map.put("account_sid", DaoUtils.writeSid(incomingPhoneNumber.getAccountSid()));
-        map.put("phone_number", incomingPhoneNumber.getPhoneNumber());
+        map.put(PHONE_NUM, incomingPhoneNumber.getPhoneNumber());
```
gvagenas: @jaimecasero why don't we move all string to constant strings?

jaimecasero: i just moved the ones i modifed, but we could potentially move all...

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
jaimecasero: ummm,interesting, i think this coming from old code, let me check what is there previously. The requirement is not clear to me, but your suggestion makes sense to me..

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
jaimecasero: here we are getting the regexes, that are expected not to be pureSIP. This is to let the rDBMS discard lots of rows to compare, since the way we get regexes is with a very awful combination of Like operators.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
jaimecasero: which rule 
-if destOrg != null -> byDestOrg
-if srcOrg !0null && srcOrg!=destOrg -> bypureSIP(false)

i think this use case is supported by next rule srcOrg!=null and srcOrg!=destOrg -> byPureSIP(false) so sipnumbers are not included in results...

Notice SMPPHandler will not provide any srcOrg or Destorg, but we still need to apply some numberSlection logic to SMPP...

gvagenas: @jaimecasero I refer to the rule that the comment is attached, which is 
```
if (sourceOrganizationSid!= null &&
                !sourceOrganizationSid.equals(destinationOrganizationSid)) {
            filterBuilder.byPureSIP(Boolean.FALSE);
        }
```

I think there is a misunderstanding here, maybe I didn't completely understand the role of NumberSelectorService.findSingleNumber() 

As far as I can understand this method is used to match a specific number only, right?

I think we absolutely need more comments and details on the NumberSelectorService

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
jaimecasero: @gvagenas  you are right, special star matching should be last attempt,even after regexes, i will change impl/test for this

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
gvagenas: @jaimecasero I am might be missing something here but the IncomingPhoneNumbers created for Regex rules will be `pureSip` numbers.
The Regex IncomingPhoneNumbers will be created using IncomingPhoneNumber REST API and providing `pureSip=true` or using Dashboard `sip number` tab

restcomm/restcomm.telephony/resources/mybatis.xml
gvagenas: @jaimecasero my comment here was just a suggestion for how to get rid the various mapper files as you suggested.
The mybatis.xml here is used for testing.
Not important now, it was just a comment

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
jaimecasero: @gvagenas umm, i expected regexes to be created as pureSIP=false, since I assumed pureSIP=true was for sipURIs numbers... let me check that again.

gvagenas: @jaimecasero You are right, my misunderstanding, I mixed up the pureSip flag role

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
jaimecasero: this particular rule was copied from previous OrganizationUtil. I can provide more comments though..

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
gvagenas: @jaimecasero the requirement here is to differentiate provider numbers (Nexmo) from the rest. For provider numbers we need to follow different flow to purchase it from the provider. Other than that all numbers are the same and pureSip flag is enough(before regex) to my opinion to cover this requirement. With the REGEX and the Smpp restrictions, we might need another flag for REGEX numbers

jaimecasero: changed to pureSIP=true

restcomm/restcomm.telephony/resources/mybatis.xml
jaimecasero: i completely removed this unused dir,and the pom part as discussed.

PR #2641, opened 2017-11-18
Issue 2620

ddhuy:
Please help review the fix and share with me your feedback.


deruelle:
@gsaslis can you please help review this one ?

deruelle:
@ddhuy can you please sign the Contributor License Agreement at https://telestax.com/open-source/#Contribute so we can accept the contribution ?

ddhuy:
@deruelle , I have signed the Contributore License Agreement.

restcomm/restcomm.identity/src/main/java/org/restcomm/connect/identity/EmailValidator.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final String VALID_EMAIL_ADDRESS_REGEX = "^[_A-Za-z0-9-\\+]+(\\.[_A-Za-z0-9-]+)*@"
+                                                          + "[A-Za-z0-9-]+(\\.[A-Za-z0-9]+)*(\\.[A-Za-z]{2,})$";
+
+    /**
+     * Verify an email address has valid format or not, use the default regex
+     *
+     * @param email: the email address need to be verified
+     * @return true if email format is valid, false if email format is invalid
+     */
+    public static boolean isValidEmailFormat ( String email ) {
```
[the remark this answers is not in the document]
ddhuy: @gsaslis I have just incorporated your comment using JavaEmail API. I agree with you that using external library is better than regex.
And I found that in class EmailMessageEndpoint we are using the regex, so should we change them?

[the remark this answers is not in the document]
gsaslis: @ddhuy yes, i would suggest we include that change as well in this PR.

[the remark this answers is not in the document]
ddhuy: @gsaslis , I changed to JavaMail as mentioned in class EmailMessageEndpoint as well. Please have a review and share with me your feedback.

restcomm/restcomm.identity/src/main/java/org/restcomm/connect/identity/EmailValidator.java
```diff
@@ -20,33 +20,33 @@
...
-        Pattern pattern = Pattern.compile(regex);
-        return pattern.matcher(email).matches();
+        boolean isValid = false;
+        try {
+            InternetAddress emailAddress = new InternetAddress(email);
+            emailAddress.validate();
+            isValid = true;
+        } catch (AddressException ex) {
+            isValid = false;
+            logger.error("Email " + email + " is invalid");
```
gsaslis: i would change the log level here to `warn` (if not `info`). 

(`error` is usually reserved for severe issues that affect the application health. on the other hand, this is very much a normal operation of handling invalid user input - `info` sounds like a better fit imho)

ddhuy: @gsaslis Agree with you about the log level. I am prefer changing it to `warn`. Thanks for your suggestion.
About the [b001a77](https://github.com/RestComm/Restcomm-Connect/commit/b001a77c8ea7f9d1757315dd5affb2e5d52b157c), I see that we had two different classes with exactly same purpose. So I think it should be better if we only have one interface to raise the email invalid format exception.

gsaslis:
@ddhuy the problem with commit https://github.com/RestComm/Restcomm-Connect/pull/2641/commits/b001a77c8ea7f9d1757315dd5affb2e5d52b157c is that you've changed the type of the `InvalidEmailException`: used to be `Exception`, now it's `RuntimeException`. 

This means that its usages (e.g. https://github.com/RestComm/Restcomm-Connect/blob/e449462909d5620daa873e9d20818fba6413ac5c/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/EmailMessagesEndpoint.java#L91 ) after your change now only catch a `RuntimeException`, instead of the more general `Exception`. 

I am not sure this is a valid / intended change... ?

ddhuy:
@gsaslis , Thanks for your reviewing time and comments. I think make the `InvalidEmailException` sub-class from `RuntimeException` is more cleaner and easier for usage. Sub-class from `Exception` or `RuntimeException` has not much different, so I will revert it back to keep minimum effect of the change.

gsaslis:
@ddhuy just to be clear: there _is_ a difference between `Exception` and `RuntimeException` and it's a *big* one. The former is a checked exception, the latter is unchecked, leading to completely different handling.

Here's some further reading on this topic https://www.ibm.com/developerworks/library/j-jtp05254/index.html , but there's a broad discussion on checked vs. unchecked exceptions and google is your friend.  ; )

I am passing your PR on for review from the core team.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
```diff
@@ -506,12 +510,22 @@ private Account prepareAccountForUpdate(final Account account, final Multivalued
                 } else
                     throw new AuthorizationException();
             }
-        } catch (AuthorizationException | AccountAlreadyClosed | PasswordTooWeak e) {
+            if (data.containsKey("EmailAddress")) {
+                String newEmailAddress = data.getFirst("EmailAddress").toLowerCase();
```
jaimecasero: any reason to force lowerCase here? AFAIK the user/local part is case sensitive form SMTP spec...

ddhuy: In `createForm` function I see the created email is always lower case so that is the reason I forced lower case here to be consistent with creating process.

ddhuy:
Hi all, I would like to take a resolve for conflicting. Please help cross check them again

gsaslis:
Thanks @ddhuy ! 

@jaimecasero / @RestComm/vvs-squad does this look good to merge?

jaimecasero:
i was thinking  email is sometimes used as account key, so im not sure making email address modifiable could have some impact. Let me think about it...

PR #2645, opened 2017-11-21
Regex may potential match all traffic

jaimecasero:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/NumberSelectorService.java
```diff
@@ -0,0 +1,325 @@
...
+     *
+     * In this case, organization details are required.
+     *
+     * @param numberQueries
+     * @param sourceOrganizationSid
+     * @param destOrg
+     * @return the longest regex matching any number in the list, null if no match
+     */
+    private IncomingPhoneNumber findByRegex(List<String> numberQueries,
+            Sid sourceOrganizationSid, Sid destOrg) {
```
ghjansen: Parameter `sourceOrganizationSid` not used in the scope.

jaimecasero: agree, we may remove this, let me quickly fix that

jaimecasero:
@ghjansen i have merged this with categories branch and the result oftesting is https://cxs.restcomm.com/view/Restcomm/job/Restcom-CI-Multibranch/job/issue1233_agg_annos/lastCompletedBuild/testReport/ -> 17 fails

im checking locally whether these are broken or just unstable

jaimecasero:
testsuite is finally positive after latest changes

PR #2662, opened 2017-11-23
Fixes typo in ClientConverter

### LINKED ISSUE #2657: Clients json converter does not return voice_application_sid
There is a bug in Clients endpoint while retrieving clients in JSON. The 'voice_application_sid' property is not present. It exists in xml version of the method though.

After doing some investigation it seems it was a bug introduced by writePushClientIdentity() writer:

https://github.com/RestComm/RestComm-Connect/blob/ea3b0ddc78f3c0b44b344d8f3f32aecfae9bc4cf/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/converter/AbstractConverter.java#L461-L461

otsakir:
Fixes bug that prevented the voice_application_sid from showing in JSON responses.

Fixes #2657


jaimecasero:
is ok to merge for me

PR #2677, opened 2017-11-30
Accounts Migration API - #Restcomm-1076

gvagenas:
PR contains the patch for account (and child accounts) migration to a new organization.
Also includes test cases and updated documentation


restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api/account-api.adoc
```diff
@@ -41,6 +41,25 @@ Accounts and sub-accounts are useful for things like segmenting phone numbers an
 |Password |A password that will be used to generate the AuthToken for the new Account resource.
 |==============================================================================================
 
+=== Account Migration to new Organization
+
```
jaimecasero: shoudl we document the different status codes (200->migrated,403->only superadmin, 400->nochildOgsuperadmin) and their meaning

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
```diff
@@ -597,6 +583,118 @@ protected Response updateAccount(final String identifier, final MultivaluedMap<S
...
+
+        //Attempt to get Organization by domain name
+        organization = organizationsDao.getOrganizationByDomainName(organizationId);
+
+        if (organization != null) {
+            return organization;
+        } else {
+            //Attempt to match organizationId to organizationSid
+            try {
+                organizationSid = new Sid(organizationId);
```
jaimecasero: Suggestion:can we do this first to check whether this is domain or Sid. If you can create a Sid form it, we know for sure its an Sid, if not, we assume is a domain (which validation is more opened). This may optimize the number of queries required....

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
```diff
@@ -597,6 +583,118 @@ protected Response updateAccount(final String identifier, final MultivaluedMap<S
...
+        Organization organization = getOrganization(data);
+        //Validation 2 - Check if data contains Organization (either SID or domain name)
+        if (organization == null) {
+            return status(PRECONDITION_FAILED).entity("Missing Organization SID or Domain Name").build();
+        }
+
+        Account operatingAccount = getOperatingAccount(identifier);
+
+        //Validation 3 - Operating Account shouldn't be null;
+        if (operatingAccount == null) {
```
jaimecasero: clarification:what is the meaning/purpose of this OperatingAccoutn check here?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
```diff
@@ -597,6 +583,118 @@ protected Response updateAccount(final String identifier, final MultivaluedMap<S
...
+            String msg = String.format("Parent Account %s migrated to Organization %s", modifiedAccount.getSid(), organization.getSid());
+            logger.debug(msg);
+        }
+
+        //Update Child accounts and their numbers
+        List<Account> childAccounts = accountsDao.getChildAccounts(operatingAccount.getSid());
+        for (Account child : childAccounts) {
+            if (!child.getOrganizationSid().equals(organization.getSid())) {
+                Account modifiedChildAccount = child.setOrganizationSid(organization.getSid());
+                accountsDao.updateAccount(modifiedChildAccount);
```
jaimecasero: is this part of the same DB transaction as all the DAO updates in this business case? how do we tackle intermidiate failures? Are we expecting failures in the middle?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsMigrationEndpointTest.java
```diff
@@ -0,0 +1,148 @@
...
+        assertEquals(newOrganizationSid, parentNewOrg);
+        assertTrue(!currentOrg.equalsIgnoreCase(parentNewOrg));
+
+        JsonObject childAccount1 = RestcommAccountsTool.getInstance().getAccount(deploymentUrl.toString(), adminUsername, adminAuthToken, childSid1);
+        assertNotNull(childAccount1);
+        String childNewOrg = childAccount1.get("organization").getAsString();
+        assertEquals(newOrganizationSid, childNewOrg);
+
+        //Migrate Parent Account and child accounts
+        assertNotNull(RestcommAccountsTool.getInstance().migrateAccount(deploymentUrl.toString(), adminUsername,
```
jaimecasero: clarification:what is the purpose of this line? why do we invoke migrate again

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsMigrationEndpointTest.java
```diff
@@ -0,0 +1,148 @@
...
+        JsonObject currentChildAccout = RestcommAccountsTool.getInstance().getAccount(deploymentUrl.toString(), adminUsername, adminAuthToken, childSid1);
+        assertNotNull(currentChildAccout);
+        String currentChildOrg = currentChildAccout.get("organization").getAsString();
+        assertEquals(originalOrganizationSid, currentChildOrg);
+
+        //Migrate Parent Account and child accounts
+        JsonObject response = RestcommAccountsTool.getInstance().migrateAccount(deploymentUrl.toString(), adminUsername,
+                adminAuthToken, childSid1, newOrganizationSid);
+
+        assertNull(response);
```
jaimecasero: can we assert status code as well?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsMigrationEndpointTest.java
```diff
@@ -0,0 +1,148 @@
...
+        JsonObject currentChildAccout = RestcommAccountsTool.getInstance().getAccount(deploymentUrl.toString(), adminUsername, adminAuthToken, childSid1);
+        assertNotNull(currentChildAccout);
+        String currentChildOrg = currentChildAccout.get("organization").getAsString();
+        assertEquals(originalOrganizationSid, currentChildOrg);
+
+        //Migrate Parent Account and child accounts
+        JsonObject response = RestcommAccountsTool.getInstance().migrateAccount(deploymentUrl.toString(), parentUsername,
+                parentAuthToken, parentSid, newOrganizationSid);
+
+        assertNull(response);
```
jaimecasero: assert status code?

jaimecasero:
waiting for testsuite results for final verification https://cxs.restcomm.com/view/Restcomm/job/Restcom-CI-Multibranch/view/change-requests/job/PR-2677/

restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api/account-api.adoc
gvagenas: @jaimecasero good point, I will add this

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
gvagenas: @jaimecasero sure I can modify this to improve the method

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
gvagenas: @jaimecasero this is a fail fast check, if the `identifier` passed in the request, doesn't match an existing account, then request should fail fast.

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
gvagenas: @jaimecasero all the DAO updates are separate DB transactions and there is no handling for failures in the middle.
What is your suggestion for this?
I would avoid to over-complicate the feature since 99% this API will only be used ONCE to move the accounts to a specific organization and never again.
Maybe we should delegate verification and failure handling to the admin/script that will use the API.

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsMigrationEndpointTest.java
gvagenas: @jaimecasero if status of the http response on the migration request from `RestcommAccountsTool.getInstance().migrateAccount()` method is 200 OK, the method response will be NOT NULL.
The execution of the migrate method again is needed to move the previously migrated accounts back to the original organization so subsequent test will work as expected

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsMigrationEndpointTest.java
gvagenas: @jaimecasero I don't think its important.

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsMigrationEndpointTest.java
gvagenas: @jaimecasero I don't think its important.

PR #2680, opened 2017-11-30
Restcomm 1350

maria-farooq:


restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
```diff
@@ -654,4 +650,108 @@ public Response deleteIncomingPhoneNumber(final String accountSid, final String
...
+        if(incomingPhoneNumbers != null) {
+            for (int i=0; i<incomingPhoneNumbers.size(); i++){
+                IncomingPhoneNumber incomingPhoneNumber = incomingPhoneNumbers.get(i);
+                incomingPhoneNumber.setOrganizationSid(destinationOrganization.getSid());
+                //update organization in db
+                dao.updateIncomingPhoneNumber(incomingPhoneNumber);
+                //update number at provider's end
+                if(!updateNumberAtPhoneNumberProvisioningManager(incomingPhoneNumber, destinationOrganization)){
+                    //if number could not be updated at provider's end, log the error and keep moving to next number.
+                    logger.error(String.format("could not update numner %s at number provider %s ", incomingPhoneNumber.getPhoneNumber(), phoneNumberProvisioningManager));
```
jaimecasero: maybe error level is too much. INFO or WARN would be enough i think...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
```diff
@@ -654,4 +650,108 @@ public Response deleteIncomingPhoneNumber(final String accountSid, final String
...
+    protected Response migrateIncomingPhoneNumbers(String targetAccountSid, MultivaluedMap<String, String> data, MediaType responseType) {
+        Account effectiveAccount = userIdentityContext.getEffectiveAccount();
+        secure(effectiveAccount, "RestComm:Modify:IncomingPhoneNumbers");
+        allowOnlySuperAdmin();
+        try{
+            Account targetAccount = accountsDao.getAccount(targetAccountSid);
+            // this is to avoid if mistakenly provided super admin account as targetAccountSid
+            // if this check is not in place and someone mistakenly provided super admin
+            // then all accounts and sub account in platform will be impacted
+            if(targetAccount.getParentSid() == null){
```
jaimecasero: do we need to check if targetAccount is null, or can we assume this will alwasy be non Null?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
maria-farooq: since this part of API is for admins, error would be something they would be interested to look at in logentries etc

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java
maria-farooq: handled

restcomm/restcomm.docs/sources-asciidoc/src/main/asciidoc/api/incoming-phone-numbers-api.adoc
```diff
@@ -578,3 +578,42 @@ match different numbers.
 
 
 
+[[resource-uri-5]]
+==== Resource URI
+
+*/2012-04-24/Accounts/\{TargetAccountSid}/IncomingPhoneNumbers/migrate*
```
jaimecasero: as in accoutn migration API, should we docuemnt possible response status? not sure this is current practice, but i expect status codes to be explained...

maria-farooq: currently adding status codes is not part of practice since mostly we use standard http codes not restcomm specific.

PR #2699, opened 2017-12-06
Pure sip provider

maria-farooq:


gvagenas:
Patch and test case looks fine to me @maria-farooq

PR #2701, opened 2017-12-06
Adjust margins for charts to show larger numbers

### LINKED ISSUE #2383: UI/Dashboard: Voice Analytics graph displays illegible number when calls surpass 6 figures numbers
As seen in following image, when calls surpass 100k calls, the number becomes illegible as the sixth figure is not shown (even the fifth one is not 100% shown). Obviously this would become worse with higher orders (more than 6 figures numbers).

image

ammendonca:
Also removed side labels and adjusted messages top margin to align with calls chart dimensions

Fixes #2383


ammendonca:
image

PR #2702, opened 2017-12-06
fix 2497: suspend account

ddhuy:
We can now suspend an account, all of its DIDs are prevented to call. Sub-accounts cannot log-in also.


gsaslis:
Thanks for the PR @ddhuy !

We would still need some tests around this to be sure we're not breaking anything critical with these changes. 

Would you please include the following test cases: 

- No REST API request is allowed for a SUSPENDED account. Account cannot login, create calls, search DID etc.
- All resources (apps, clients, DIDs etc) are suspended and cannot be used. Clients cannot be registered, DID numbers return 403
- All sub-accounts of the account and their resources are also suspended and cannot be used.
- The Account, Sub-Accounts and all related resources still exist / are not removed from db.
- A suspended account can move to either CLOSED state or to ACTIVE state again.
- Moving to ACTIVE state means that Account and sub-accounts are active and can be used again, same for all related resources.
- Moving to CLOSED state means that Account, sub-accounts and related resources are closed forever and will be never used again. External resources like, purchased DID number will be released (so no more costs) but everything else (account, sub-account, resources) will remain in the DB for audit purposes.

Looking forward to reviewing once these are in place!

ddhuy:
@gsaslis I have added regression tests as your suggestion. Please help verify them again.

ddhuy:
This PR is so far out of date with master. So I create a new [PR 2904](https://github.com/RestComm/Restcomm-Connect/pull/2904) one.

gsaslis:
thanks @ddhuy !

PR #2705, opened 2017-12-07
PushNotificationServerHelper 'httpPost.releaseConnection() added

agafox:


restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/push/PushNotificationServerHelper.java
```diff
@@ -94,6 +94,8 @@ public Void call() throws Exception {
                     }
                 } catch (Exception e) {
                     logger.error("Exception while sending push server notification, " + e);
+                } finally {
+                    httpPost.releaseConnection();
```
jaimecasero: we are generally using response.close instead in other scenarios, but if request fails with no response (timeoute,remoteexception..) i guess this req.releaseConnection is better actually.

Another concern, not applicable in this case, is that response content should be consumed asap after response, when no expcetion is thrown. I guess after req.releaseConnection reading response content will not be possible. Not sure if this code is still missing the response content consumption (    [EntityUtils.consume(entity1);](https://hc.apache.org/httpcomponents-client-ga/quickstart.html))..., can you check that ? and after that we can create a story to cover all http client usages in the RC applciation with clear guidelines.

jaimecasero:
as per olegs explanantion releaseConnection will free response content resources as well...

PR #2720, opened 2017-12-18
Fix for intermittence of parameter https.protocols from JAVA_OPTS when starting and stopping docker containers

ghjansen:
JIRA: https://telestax.atlassian.net/browse/RESTCOMM-1327

### DIFF
--- a/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh
+++ b/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh
@@ -637,12 +637,12 @@ otherRestCommConf(){
 	mv $FILE.bak $FILE
 
     #Remove if is set in earlier run.
-    grep -q 'allowLegacyHelloMessages' $RESTCOMM_BIN/standalone.conf && sed -i "s|-Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=.* ||" $RESTCOMM_BIN/standalone.conf
+    grep -q 'allowLegacyHelloMessages' $RESTCOMM_BIN/standalone.conf && sed -i -E "s/(.*)( -Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=)(true|false)(.*)/\1\4/" $RESTCOMM_BIN/standalone.conf
 
     if [[ "$SSLSNI" == "false" || "$SSLSNI" == "FALSE" ]]; then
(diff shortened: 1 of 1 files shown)


restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh
```diff
@@ -637,12 +637,12 @@ otherRestCommConf(){
...
 
     #Remove if is set in earlier run.
-    grep -q 'allowLegacyHelloMessages' $RESTCOMM_BIN/standalone.conf && sed -i "s|-Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=.* ||" $RESTCOMM_BIN/standalone.conf
+    grep -q 'allowLegacyHelloMessages' $RESTCOMM_BIN/standalone.conf && sed -i -E "s/(.*)( -Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=)(true|false)(.*)/\1\4/" $RESTCOMM_BIN/standalone.conf
 
     if [[ "$SSLSNI" == "false" || "$SSLSNI" == "FALSE" ]]; then
-		  sed -i "s|-Djava.awt.headless=true|& -Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=false |" $RESTCOMM_BIN/standalone.conf
+		  sed -i "s|-Djava.awt.headless=true|& -Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=false|" $RESTCOMM_BIN/standalone.conf
 	else
-	 	  sed -i "s|-Djava.awt.headless=true|& -Dsun.security.ssl.allowLegacyHelloMessages=false -Djsse.enableSNIExtension=true |" $RESTCOMM_BIN/standalone.conf
```
jaimecasero: i guess this space removal is key to the wildcard matching that happens later. i wonder how we could mark this as important, a comment...

ghjansen: Not exactly @jaimecasero , i did some grooming on white spaces just because they were accumulating on each start/stop. So after a while `standalone.conf` file were found with sth like...
```
... -Djava.awt.headless=true                        -Dhttps.protocols=TLSv1.1,TLSv1.2 ...
```
...that would keep growing between parameters, but now parameters are added and removed continuously preserving single white spaces between them. JVM doesn't care about this since it seems to remove exceedent white spaces at runtime, so i think its just an improvement.

PR #2723, opened 2017-12-19
Restcomm 1259

maria-farooq:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -2749,9 +2766,15 @@ public void execute(final Object message) throws Exception {
                             logger.info("Canceled branch: " + branch.path()+", isTerminated: "+branch.isTerminated());
                         }
                     }
-                    // Stop playing the ringing tone from inbound call
-                    msResponsePending = true;
-                    call.tell(new StopMediaGroup(), self());
+                    // if enable200OkDelay is true, call was not playing ringing tone, so no need to send stop ringing
+                    if(enable200OkDelay){
+                        outboundCallResponse = SipServletResponse.SC_REQUEST_TIMEOUT;
+                        call.tell(new StopWaiting(), self());
```
gvagenas: @maria-farooq I think we shouldn't ask the initial call actor to stop waiting and answer the call at this point because there is a chance there will be no next verb (either subsequent verb or verb from Dial action).

To my point of view, we should first check if there is next verb (either by parser getNext or by executing the Dial Action) and if the next verb is Say, Play, Gather or other verb that involve Media Server, only then ask the call to answer. 

Otherwise, we can terminate the original call. Which also means that verbs that don't need Media Server will be executed just fine, such as Sms, Email etc.

maria-farooq: i see the point, will update the patch so we only answer the call if we have relevant upcoming tag

maria-farooq:
@gvagenas  All delay related test clases are green so if you have reviewed the patch we are ok to merge

PR #2751, opened 2018-01-08
Fix mediaserver connectivity issues on docker compose

gsaslis:
On macOS (at least) the SIP client fails to connect to the Restcomm Media Server, when running with `docker compose up` as the IP address the RMS scripts pick up is the docker one (e.g. `172.24.0.2`) and that's not accessible from the host. 

The extra env var allows setting a routable IP address which a SIP client can use to connect to both Restcomm Connect, as well as the Media Server. 

A special thanks to @hrosa for the help debugging this! <3

/cc @agafox: I believe this still doesn't work with `127.0.0.1` but I think this brings us a step closer.


deruelle:
@agafox @hrosa did you confirm that works ? Can this be merged ?

gsaslis:
@deruelle  I think we should go ahead and merge this in,  as it seems to be affecting a few people and doesn’t really affect the application itself  to warrant a full review from the maintainers.

PR #2757, opened 2018-01-16
Restcomm 1512 - Extensions API changes needed for FAC

gvagenas:
Changes needed to the Extensions API to support Feature Access Control feature

### DIFF
--- a/restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/IExtensionFeatureAccessRequest.java
+++ b/restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/IExtensionFeatureAccessRequest.java
@@ -1,0 +1,14 @@
+package org.restcomm.connect.extension.api;
+
+import org.restcomm.connect.commons.dao.Sid;
+
+public interface IExtensionFeatureAccessRequest extends IExtensionRequest{
+
+    void setAccountSid (Sid accountSid);
+
+    String getDestinationNumber();
+
+    void setDestinationNumber(String destinationNumber);
+
+
+}
--- a/restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/IExtensionFeatureAccessRequest.java
+++ b/restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/IExtensionFeatureAccessRequest.java
@@ -1,3 +1,23 @@
+/*
+ * TeleStax, Open Source Cloud Communications
+ * Copyright 2011-2013, Telestax Inc and individual contributors
+ * by the @authors tag.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
 package org.restcomm.connect.extension.api;
 
 import org.restcomm.connect.commons.dao.Sid;
--- a/restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/RestcommExtensionGeneric.java
+++ b/restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/RestcommExtensionGeneric.java
@@ -21,7 +21,6 @@
 package org.restcomm.connect.extension.api;
 
 import javax.servlet.ServletContext;
-import javax.servlet.sip.SipServletRequest;
 
 /**
  * @author <a href="mailto:gvagenas@gmail.com">gvagenas</a>
@@ -43,14 +42,14 @@ public interface RestcommExtensionGeneric {
      * and either block/allow or modify the session before Restcomm process it
      * @return ExtensionResponse see ExtensionResponse
      */
-    ExtensionResponse preInboundAction(SipServletRequest request);
+    ExtensionResponse preInboundAction(IExtensionRequest extensionRequest);
     /**
      * Method that will be executed AFTER the process of an Incoming session
      * Implement this method so you will be able to check the Incoming session
      * and either block or allow or modify the session after Restcomm process it
      * @return ExtensionResponse see ExtensionResponse
      */
-    ExtensionResponse postInboundAction(SipServletRequest request);
+    ExtensionResponse postInboundAction(IExtensionRequest extensionRequest);
     /**
      * Method that will be executed before the process of an Outbound session
      * Implement this method so you will be able to check the Outbound session
--- a/restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
+++ b/restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
@@ -1,5 +1,6 @@
 package org.restcomm.connect.extension.controller;
 
+import org.restcomm.connect.extension.api.ApiRequest;
 import org.restcomm.connect.extension.api.ExtensionResponse;
 import org.restcomm.connect.extension.api.ExtensionType;
 import org.restcomm.connect.extension.api.IExtensionRequest;
@@ -121,7 +122,6 @@ public class ExtensionController {
         ExtensionResponse response = new ExtensionResponse();
 //        response.setAllowed(true);
         if (extensions != null && extensions.size() > 0) {
-
             for (RestcommExtensionGeneric extension : extensions) {
                 if(logger.isInfoEnabled()) {
                     logger.info( extension.getName()+" is enabled="+extension.isEnabled());
@@ -139,6 +139,31 @@ public class ExtensionController {
     }
 
     public ExtensionResponse executePostInboundAction(final IExtensionRequest er,  List<RestcommExtensionGeneric> extensions) {
+        ExtensionResponse response = new ExtensionResponse();
+        //TODO: implement actual calls
+        return response;
+    }
+
+    public ExtensionResponse executePreApiAction(final ApiRequest apiRequest, List<RestcommExtensionGeneric> extensions) {
+        ExtensionResponse response = new ExtensionResponse();
+
+        if (extensions != null && extensions.size() > 0) {
+            for (RestcommExtensionGeneric extension : extensions) {
+                if(logger.isInfoEnabled()) {
+                    logger.info( extension.getName()+" is enabled="+extension.isEnabled());
+                }
+                if (extension.isEnabled()) {
+                    response = extension.preApiAction(apiRequest);
+                    //fail fast
+                    if (!response.isAllowed())
+                        break;
+                }
+            }
+        }
+        return response;
+    }
+
+    public ExtensionResponse executePostApiAction(final ApiRequest apiRequest, List<RestcommExtensionGeneric> extensions) {
         ExtensionResponse response = new ExtensionResponse();
         //TODO: implement actual calls
         return response;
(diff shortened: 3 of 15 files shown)


restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/RestcommExtensionGeneric.java
```diff
@@ -43,14 +42,14 @@
      * and either block/allow or modify the session before Restcomm process it
      * @return ExtensionResponse see ExtensionResponse
      */
-    ExtensionResponse preInboundAction(SipServletRequest request);
+    ExtensionResponse preInboundAction(IExtensionRequest extensionRequest);
```
jaimecasero: i like the idea of being independent to underlying framework (servlet,akka), so good choice on having our own request. It would be still a good idea to be able to recover ref to original framework objects using generic container of objects, so extension impl may use them if required. I noticed in the "init" method we still inject a ServletContext, so the extension is still tied to specific ServletAPI. not sure if is time to have an specific ExtensionContext as well, and pass it as part of the request...

restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
```diff
@@ -116,4 +117,55 @@ public ExtensionResponse executePostOutboundAction(final IExtensionRequest er, L
         //TODO: implement actual calls
         return response;
     }
+
+    public ExtensionResponse executePreInboundAction(final IExtensionRequest er, List<RestcommExtensionGeneric> extensions) {
+        ExtensionResponse response = new ExtensionResponse();
```
jaimecasero: what is the meaning of this default response?, how do we detect if this response is coming from an actual extension or not? maybe the empty response is good enough for the framework, just wondering...

restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
```diff
@@ -116,4 +117,55 @@ public ExtensionResponse executePostOutboundAction(final IExtensionRequest er, L
...
+                    if (!response.isAllowed()){
+                        break;
+                    }
+                }
+            }
+        }
+        return response;
+    }
+
+    public ExtensionResponse executePostInboundAction(final IExtensionRequest er,  List<RestcommExtensionGeneric> extensions) {
```
jaimecasero: this extension API is starting to feel very specific ,with many variants of same purpose. Not sure if we could push all this executePre/PostXAction methods into the request interface, so the request tells the type of exec point X by a using a getter in the req interface... I think we could have a single pre and post methods...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
```diff
@@ -376,57 +378,73 @@ protected Response putAccount(final MultivaluedMap<String, String> data, final M
...
-            return status(BAD_REQUEST).entity(exception.getMessage()).build();
-        } catch (PasswordTooWeak passwordTooWeak) {
-            return status(BAD_REQUEST).entity(buildErrorResponseBody("Password too weak",responseType)).type(responseType).build();
-        }
 
-        // If Account already exists don't add it again
+        ExtensionController ec = ExtensionController.getInstance();
+        ApiRequest apiRequest = new ApiRequest(sid.toString(), data, ApiRequest.Type.CREATE_SUBACCOUNT);
+
+        if (executePostApiAction(apiRequest)) {
```
jaimecasero: is this really post,or should be pre?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/BaseVoiceInterpreter.java
```diff
@@ -1564,6 +1575,23 @@ public void execute(final Object message) throws Exception {
...
+                    if (logger.isDebugEnabled()) {
+                        final String errMsg = "ASR feature is not allowed";
+                        logger.debug(errMsg);
+                    }
+                    String errMsg = "ASR feature is not allowed";
+                    notification(WARNING_NOTIFICATION, 11001, errMsg);
+                    call.tell(new Hangup(errMsg), source);
+                    return;
+                }
+                ec.executePostOutboundAction(er, extensions);
```
jaimecasero: should we delay the invocation of PostAction to when the ASR is actually executed?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/SmsInterpreter.java
```diff
@@ -378,7 +388,24 @@ public void onReceive(final Object message) throws Exception {
...
+                    fsm.transition(message, creatingSmsSession);
+                } else {
+                    if (logger.isDebugEnabled()) {
+                        final String errMsg = "Outbound SMS is not Allowed";
+                        logger.debug(errMsg);
+                    }
+                    final NotificationsDao notifications = storage.getNotificationsDao();
+                    final Notification notification = notification(WARNING_NOTIFICATION, 11001, "Outbound SMS is now allowed");
+                    notifications.addNotification(notification);
+                    fsm.transition(message, finished);
```
jaimecasero: are we covering postaction invocation later as part of FSM?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -998,7 +1000,25 @@ private void onTagMessage(Object message) throws TransitionFailedException, Tran
...
+                if (er.isAllowed()) {
+                    fsm.transition(message, creatingSmsSession);
+                } else {
+                    if (logger.isDebugEnabled()) {
+                        final String errMsg = "Outbound SMS is not Allowed";
+                        logger.debug(errMsg);
+                    }
+                    final NotificationsDao notifications = storage.getNotificationsDao();
+                    final Notification notification = notification(WARNING_NOTIFICATION, 11001, "Outbound SMS is now allowed");
+                    notifications.addNotification(notification);
```
jaimecasero: are we covering postaction invocation as part of fsm lateR?

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
```diff
@@ -228,45 +245,65 @@ private boolean redirectToHostedVoiceApp(final ActorRef self, final SipServletRe
...
+                        final String errMsg = "Inbound USSD session is not Allowed";
+                        logger.debug(errMsg);
+                    }
+                    String errMsg = "Inbound USSD session to Number: " + number.getPhoneNumber()
+                            + " is not allowed";
+
+                    sendNotification(errMsg, 11001, "warning", true);
+                    final SipServletResponse resp = request.createResponse(SC_FORBIDDEN, "Inbound USSD session is not Allowed");
+                    resp.send();
+                    ec.executePostOutboundAction(far, extensions);
```
jaimecasero: are we invoking postAction here if feature isAllowed in the other branch?

restcomm/restcomm.sms/src/main/java/org/restcomm/connect/sms/SmsService.java
```diff
@@ -254,45 +256,61 @@ public void run() {
...
+                } else {
+                    if (logger.isDebugEnabled()) {
+                        final String errMsg = "Outbound SMS from Client " + client.getFriendlyName() + " not Allowed";
+                        logger.debug(errMsg);
+                    }
+                    String errMsg = "Outbound SMS from Client: " + client.getFriendlyName()
+                            + " is not Allowed";
+                    sendNotification(errMsg, 11001, "warning", true);
+                    final SipServletResponse resp = request.createResponse(SC_FORBIDDEN, "Call not allowed");
+                    resp.send();
```
jaimecasero: where do we cover postAction in this execution point (for both allowed and notAllowed options)?

maria-farooq:
my comment on the topic is related to `RestcommExtensionGeneric.java` I think for me 
```
postApiAction
postInboundAction
postOutboundAction
preApiAction
preInboundAction
preOutboundAction
```
are same method and should have similar approach on applying extentions to an `Action` regardless of it being `pre` or `post`, `inbound` or `out bound`, `api request` or `anyother request`

restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/RestcommExtensionGeneric.java
gvagenas: ExtensionContext sounds a good idea @jaimecasero. Not sure also if its the proper time to do it. 
I think that since now the team will be more familiar with the Extensions API, for the next task we can plan and estimate enhancements like the one you suggest

restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
gvagenas: @jaimecasero The `ExtensionResponse` object will be set by each extension. The `response` holds an object that can be used to pass data from one extension back to RC, such as custom headers, custom proxy headers etc.

Currently, we don't use the `post-*-action` methods so we just create a `response` object and return it.

The idea for the `post-*-action` is to use them for maintenance and logistic, such as update logs, update db, notify services (charging) etc.....

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
gvagenas: @jaimecasero good catch, this must be `pre`. I will fix that

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/BaseVoiceInterpreter.java
gvagenas: @jaimecasero, in general, the `post` action should be executed after the `pre` action. 

At the point of your comment,  the call is already answered and maybe had already executed some RCML verbs, the interpreter started executions of the `gather` verb with `asr` which account was not permitted and thus interpreter hangup the call. I think this is the proper place for `post` action method.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/SmsInterpreter.java
gvagenas: @jaimecasero no, I missed the `post` action and I will add it now

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
gvagenas: @jaimecasero no, I missed the post action and I will add it now

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
gvagenas: @jaimecasero we need to invoke `post` action in any case, I will fix that

restcomm/restcomm.sms/src/main/java/org/restcomm/connect/sms/SmsService.java
gvagenas: @jaimecasero I missed that, I will fix this

gvagenas:
@maria-farooq @jaimecasero about your comments on improvements for Extensions API, I think those should definitely be discussed but are out of the scope of this feature and this sprint.

Any changes in the API will have an impact on existing extensions. As I suggested earlier, we should record any suggestion, discuss them and in the next sprints plan them

restcomm/restcomm.extension.api/src/main/java/org/restcomm/connect/extension/api/IExtensionFeatureAccessRequest.java
```diff
@@ -0,0 +1,14 @@
+package org.restcomm.connect.extension.api;
```
maria-farooq: Should we add copyright statement in begenning of this file?

restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/FeatureAccessRequest.java
```diff
@@ -0,0 +1,79 @@
+package org.restcomm.connect.telephony.api;
```
maria-farooq: shoudl we adde copyright statement in begening of this file

PR #2763, opened 2018-01-18
Profile Associations DAO Impl

maria-farooq:


gvagenas:
I provided fixes for maven checkstyle plugin that prevented project to properly compile

PR #2766, opened 2018-01-22
RESTCOMM-1625: Handle Route Header for MultiProvider config

abdulazizali77:
@gvagenas @deruelle George, VVS, please have a look

Motivation:
- Push routes defined in MultiProvider outbound-voice config.

Modification:
- Move the duplicated CallManager and Call addHeadersToMessage to B2BUAHelper
- Add explicit handling for Route headers
- Save outboundHeaders to incoming request SipSession and modify B2BUA invite
- Fetch outboundHeaders during challenge response and modify linked B2BU invite
- Fix proxyPassword

Result:
- WebRTC/B2BUA calls will be able to modify headers and add Route headers

### DIFF
--- a/restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/util/B2BUAHelper.java
+++ b/restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/util/B2BUAHelper.java
@@ -18,12 +18,14 @@
  *
  */package org.restcomm.connect.telephony.api.util;
 
- import java.io.IOException;
+import java.io.IOException;
 import java.math.BigDecimal;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
 import java.util.Currency;
+import java.util.Map;
 import java.util.Vector;
 
 import javax.sdp.Connection;
@@ -70,6 +72,7 @@ import org.restcomm.connect.telephony.api.CallStateChanged;
      public static final String B2BUA_LAST_REQUEST = "lastRequest";
      public static final String B2BUA_LAST_RESPONSE = "lastResponse";
      public static final String B2BUA_LAST_FINAL_RESPONSE = "lastFinalResponse";
+     public static final String EXTENSION_HEADERS = "extensionHeaders";
      private static final String B2BUA_LINKED_SESSION = "linkedSession";
      private static final String CDR_SID = "callDetailRecord_sid";
 
@@ -302,6 +305,9 @@ import org.restcomm.connect.telephony.api.CallStateChanged;
                      ((SipSessionExt) outRequest.getSession()).setBypassProxy(true);
                  }
              }
+
+             Map<String,ArrayList<String>> extensionHeaders = (Map<String,ArrayList<String>>)incomingSession.getAttribute(EXTENSION_HEADERS);
+             addHeadersToMessage(outRequest, extensionHeaders, sipFactory);
              outRequest.send();
              Address originalFromAddress = request.getFrom();
              SipURI originalFromUri = (SipURI) originalFromAddress.getURI();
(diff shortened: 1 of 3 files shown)


deruelle:
adding @leftyb as well.

restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/util/B2BUAHelper.java
```diff
@@ -626,4 +632,120 @@ public static boolean isB2BUASession(SipServletMessage sipMessage) {
          return (linkedB2BUASession != null);
      }
 
+     /**
```
jaimecasero: this is just relocating the addHeaders logic to the B2BUAHelper to be used from anywhere as an util,right?

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
```diff
@@ -702,72 +703,6 @@ public void run() {
...
     }
 
-    /**
-     * FIXME: duplicated code make into static function or something more optimized
-     * Replace headers
-     *
-     * @param SipServletRequest message
-     * @param Map<String,       ArrayList<String> > headers
-     */
-    private void addHeadersToMessage(SipServletRequest message, Map<String, ArrayList<String>> headers) {
```
jaimecasero: i take this was duplicated from Call logic, exactly same code?

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
```diff
@@ -2272,7 +2207,8 @@ public void response(final Object message) throws IOException {
 
         // if this response is coming from a client that is in a p2p session with another registered client
         // we will just proxy the response
-        if (B2BUAHelper.isB2BUASession(response)) {
+        SipSession linkedB2BUASession = B2BUAHelper.getLinkedSession(response);
+        if (linkedB2BUASession!=null) {
```
jaimecasero: i take this is actually the only change to fix the issue

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
```diff
@@ -2289,6 +2225,9 @@ public void response(final Object message) throws IOException {
                 SipServletRequest invite = response.getRequest();
                 challengeRequest.setContent(invite.getContent(), invite.getContentType());
                 invite = challengeRequest;
+
+                Map<String,ArrayList<String>> extensionHeaders = (Map<String,ArrayList<String>>)linkedB2BUASession.getAttribute(B2BUAHelper.EXTENSION_HEADERS);
+                B2BUAHelper.addHeadersToMessage(challengeRequest, extensionHeaders, sipFactory);
```
jaimecasero: i take this is actually the only change to fix the issue, is it not the same condition as before?

restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/util/B2BUAHelper.java
abdulazizali77: Correct. We also added the specific `Route` handling to push the route

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
abdulazizali77: The only difference being the two were referring to private `sipFactory`, which we added as a param in the relocated one

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
abdulazizali77: Previously we never modified the headers for this B2BUA/challenge request case.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
abdulazizali77: `B2BUAHelper.isB2BUASession(response)` basically checks for the existence of a session in the message. 
Since the block previously did not need to access any session data, it did not retrieve the `SipSession`. Now we do need to one access session attribute `EXTENSION_HEADERS` so we do need the local `SipSession`

jaimecasero:
for me this PR is ok to be merged, we just need to confirm on testsuite results.

would it make sense to add a testcase for the new feature, to check extensionHeaders are added appropiately?

abdulazizali77:
@jaimecasero The tests PR are here
https://bitbucket.org/telestax/telscale-restcomm/pull-requests/60/add-multiprovider-route-tests

jaimecasero:
im ok merging this when test results are ready

PR #2775, opened 2018-01-25
Restcomm 1618 - Protect ExtensionController action methods for NPE

gvagenas:
Includes fixes for Extensions API execution points

### DIFF
--- a/restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
+++ b/restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
@@ -22,12 +22,14 @@ public class ExtensionController {
     private List smsSessionExtensions;
     private List ussdCallManagerExtensions;
     private List restApiExtensions;
+    private List featureAccessControlExtensions;
 
     private ExtensionController(){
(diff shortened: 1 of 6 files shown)


restcomm/restcomm.extension.controller/src/main/java/org/restcomm/connect/extension/controller/ExtensionController.java
```diff
@@ -96,41 +106,79 @@ public ExtensionResponse executePreOutboundAction(final IExtensionRequest ier, L
...
+                    try {
+                        ExtensionResponse tempResponse = extension.preOutboundAction(ier);
+                        if (tempResponse != null) {
+                            response = tempResponse;
+                            //fail fast
+                            if (!tempResponse.isAllowed()) {
+                                break;
+                            }
+                        }
+                    } catch (Exception e) {
```
jaimecasero: do we want to catch just Exception? should we try to catch Throwable? im not sure, just wondering...
My perspective is Extensions are out of RC core, and RC core works as the platform for specific Extensions. Being the ExtensionController the border from RC to extension, it may make sense to catch Throwable instead,and provide a default recovery path..

gvagenas: Changed to catch a Throwable.

PR #2781, opened 2018-01-26
fixes https://github.com/RestComm/Restcomm-Connect/issues/2779

jaimecasero:
invoke exec point only when a number has been selected


gvagenas:
Good catch @jaimecasero . The patch looks good to me and I will merge it with master

PR #2792, opened 2018-01-29
Restcomm 1622 - Patch for Dial timeout on forking

gvagenas:


restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Call.java
```diff
@@ -1023,8 +1023,17 @@ public void execute(Object message) throws Exception {
             // Send the invite.
             invite.send();
             // Set the timeout period.
-            final UntypedActorContext context = getContext();
-            context.setReceiveTimeout(Duration.create(timeout, TimeUnit.SECONDS));
+
+//            final UntypedActorContext context = getContext();
```
jaimecasero: apparently the patch is just commenting these lines...
I wonder where the part of reseting teh akka timer only on particualr message is, the one we discussed offline...

abdulazizali77: without the deadline, the `timeout` param on the `CallsEndpoint` and in the constructor is unused, 
will the `Call.onReceiveTimeout` be called with an akka timeout default?

PR #2793, opened 2018-01-29
Restcomm 1623

maria-farooq:


maria-farooq:
tests are ok now, PR can be merged

PR #2797, opened 2018-01-30
Restcomm 1633 - Don't execute RCML Dial after completed Dial in a sequential dialling scenario

gvagenas:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -975,17 +975,17 @@ private void onTagMessage(Object message) throws TransitionFailedException, Tran
                 fsm.transition(message, startDialing);
             } else if (Verbs.fax.equals(verb.name())) {
                 fsm.transition(message, caching);
-            } else if (Verbs.play.equals(verb.name())) {
+            } else if (Verbs.play.equals(verb.name()) && callState != CallStateChanged.State.COMPLETED) {
```
jaimecasero: it seems we are preventing Verb execution in COMPLETED state sparringly (not all verbs have this protection), and out of FSM transition logic.
Why dont all the verbs have same condition?
Could we prevent same thing with actual FSM configuration (adding/removign transition)?

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/IvrEndpoint.java
```diff
@@ -119,7 +119,7 @@ private void stop(Object message) {
         request.setNotifiedEntity(agent);
 
 //        //Ask RMS to notify for Recording Signal
-        if (se.getEvent().getName().equalsIgnoreCase("pr")) {
+        if (se.getEvent().getName().equalsIgnoreCase("pr") || se.getEvent().getName().equalsIgnoreCase("pa") ) {
```
jaimecasero: this is similar to another PR we just introduced. Is there any other place that needs to be updated in similar way (supporting the "pa" event)?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -2851,31 +2851,9 @@ public void execute(final Object message) throws Exception {
...
                     } else {
-                        // At this point !sender.equal(call)
-                        // Ask the parser for the next action to take.
-                        Attribute attribute = null;
-                        if (verb != null) {
-                            attribute = verb.attribute("action");
+                        if (callState == CallStateChanged.State.IN_PROGRESS) {
+                            call.tell(new Hangup(), self());
                         }
-                        if (attribute == null) {
```
jaimecasero: it seems we removed quite a few lines here, where is all this stuff happening now?

I guess this is related to previous changes on the new COMPLETED condition to execute verbs.

Still is not clear to me the relationship? Could you elaborate the high level reasoning of the PR interms of the VI FSM? what the key changes in the states/transitions? what is the purpose of this change. All this explanation is just to faciliate myself understanding of the change(hopefully will help others as well)

jaimecasero: ive seen this note that partly answer my question:

> 

Note: RCML after Dial verb should still be executed (given there is no Dial Action attribute) but not for Dial, Play, Say, Gather, Record since the call is competed. SMS, Fax, Email should be executed ok.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
jaimecasero: so, the issue is that this decision is not based on VI own state, but on the parituclar call state, hence the usage of if statements rather than FSM transitions, right?

jaimecasero: Would be nice to modularize which verbs are meant to be executed if call.COMPLETED or not?
maybe a private method supported by a Map, if the verb is container in the MAP proceed,if not,prevent. The name of the method may help to understand what is going on

suggestion : "isExecutionAllowedWhenCompleted"

the private method may introduce some debug logs to indicate if teh verb was allowed to be executed or not in addition.

gvagenas: @jaimecasero you answered correctly all the questions and you expressed the same concerns I had.

Some verbs such as SMS, Fax, Email, can be executed even when the call is completed but other not.

Also, I was also thinking that this is not an elegant way to control which verb should be executed depending on the Call state but due to the urgency of the issue and the fact that FSM refactoring is on the plan, I decided to go the fast way and wait for the CI results.

restcomm/restcomm.mgcp/src/main/java/org/restcomm/connect/mgcp/IvrEndpoint.java
gvagenas: @jaimecasero from the discussion I had with Henrique this would be enough, but I will double check with him

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
gvagenas: @jaimecasero The FSM will reach FinishDialing line 2854 https://github.com/RestComm/Restcomm-Connect/blob/21ff4940369a337f051b1ae2552a228e50b70a4b/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java#L2854 when called client sends BYE. Before we were either executing Dial Action to get new RCML or were asking parser for next RCML but we didn't disconnected caller.
That means that in a sequential RMCL scenario:
1. A was connected to B, 
2. B was sending BYE
3. A was not disconnected but instead VI was executing next RCML verb
4. Thus, next Dial verb was executed, C was ringing

Now the patch disconnects A. When A completes, will send CallStateChanged.Completed to VI which will cause FSM to move to FinishDialing again, but this time will cleanup A call actor and will check next RCML.

Since CallState is Completed, subsequent Dial verbs will not be executed because of the changes in the `onTag()` method.

maria-farooq:
just scheduled another ci build for it

maria-farooq:
we should wait for ci results for this one `https://cxs.restcomm.com/job/Restcom-CI-Multibranch/view/change-requests/job/PR-2797/3`

PR #2799, opened 2018-01-31
fixes #RESTCOMM-1635

maria-farooq:


restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/UssdPushEndpoint.java
```diff
@@ -120,12 +120,13 @@ protected Response putCall(final String accountSid, final MultivaluedMap<String,
         final String to = data.getFirst("To");
         final String username = data.getFirst("Username");
         final String password = data.getFirst("Password");
-        final Integer timeout = getTimeout(data);
+        Integer timeout = getTimeout(data);
+        timeout = timeout != null ? timeout : 30;
```
jaimecasero: is this default 30 seconds delay used elsewhere? where is the RCML counterpart defiend? shouldnt come from conf?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -475,6 +477,7 @@ public VoiceInterpreter(VoiceInterpreterParams params) {
         this.asImsUa = params.isAsImsUa();
         this.imsUaLogin = params.getImsUaLogin();
         this.imsUaPassword = params.getImsUaPassword();
+        this.timeout = params.getTimeout();
```
jaimecasero: this is maybe for later refactoring, why dont we save the ref to params directly, instead of replcaiting all params atts in VI?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
```diff
@@ -363,7 +363,8 @@ protected Response putCall(final String accountSid, final MultivaluedMap<String,
         final String to = data.getFirst("To").trim();
         final String username = data.getFirst("Username");
         final String password = data.getFirst("Password");
-        final Integer timeout = getTimeout(data);
+        Integer timeout = getTimeout(data);
+        timeout = timeout != null ? timeout : 30;
```
jaimecasero: i really would love to see this default timeout defined as a constant maybe in VI, or coming from conf...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/UssdPushEndpoint.java
maria-farooq: @jaimecasero i am improvising of existing implmentations, you can check this default 30 sec being used at following line 128.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -795,7 +798,7 @@ private void onReceiveTimeout(Object message) throws TransitionFailedException,
             fsm.transition(message, finishDialing);
         } else if (is(bridging)) {
             fsm.transition(message, finishDialing);
-        } else if (is(playing)) {
+        } else if (is(playing) || is(initializingCall)) {
```
jaimecasero: is this a new transition we didnt covered? which scenario are we covering here (timeout during initializing)?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
maria-farooq: again in this PR, its just being set as per older implementation, i just set it here instead of multiple conditional statements bellow

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
maria-farooq: during `initializingCall` we are starting a new timer, so if timer expires and we are still in `initializingCall ` phase then VI will move onto `Finished` state. JFYI we already do the same for Busy state.

maria-farooq: during `initializingCall` we are starting a new timer, so if timer expires and we are still in `initializingCall ` phase then VI will move onto `Finished` state. JFYI we already do the same for Busy state.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -1761,6 +1764,9 @@ public void execute(final Object message) throws Exception {
 
                 // Start dialing.
                 call.tell(new Dial(), source);
+                // Set the timeout period.
+                final UntypedActorContext context = getContext();
+                context.setReceiveTimeout(Duration.create(timeout, TimeUnit.SECONDS));
```
jaimecasero: @gvagenas  told me about this timer being reset in a message is received. But we may need to keep the timer even if some message is received... I found this in Akka doc https://doc.akka.io/japi/akka/current/akka/actor/NotInfluenceReceiveTimeout.html to mark messages as whether they should reset receiveTimeout or not.

Honestly i think we should not use receiveTimeout for any kind of FSM timer, as most of the times is not really related to FSM/Actor inactivity, but other kind of logic (whether we receive the proper events in proper order,and FSM transitions to diff states accordingly). Maybe we need to use another timer infra to send timeout events to the actors.  I notice this more general purpose scheduler in akka https://doc.akka.io/docs/akka/2.5/scheduler.html, not sure is available in our akka version

PR #2804, opened 2018-02-05
Restcomm 1648 - Check Account state for Telco traffic

gvagenas:


restcomm/restcomm.sms/src/main/java/org/restcomm/connect/sms/SmsService.java
```diff
@@ -169,19 +171,64 @@ private void message(final Object message) throws IOException {
...
+            }
+        }
+
         // Make sure we force clients to authenticate.
         if (client != null) {
+
+            Account clientAccount = accounts.getAccount(client.getAccountSid());
+            if (!clientAccount.getStatus().equals(Account.Status.ACTIVE)) {
+                //reject SMS since the Number belongs to an an account which is not ACTIVE
+                final SipServletResponse response = request.createResponse(SC_NOT_ACCEPTABLE);
```
jaimecasero: any reason to use 406/NOT_ACCEPTABLE? AFAIK this is more related to Accep header and contentNegotiation. what about 403/FORBIDDEN? 403 is a general purpose status to reject requests "The server understood the request, but is refusing to fulfill it" (doesnt have to be related to authorization alone). 

Also consider to put a reason phrase in addition to the code https://docs.oracle.com/cd/E13209_01/wlcp/wlss40/javadoc/jsr289/index.html?javax/servlet/sip/SipServlet.html like "Account is not Active"

restcomm/restcomm.sms/src/main/java/org/restcomm/connect/sms/SmsService.java
```diff
@@ -169,19 +171,64 @@ private void message(final Object message) throws IOException {
...
+
         final Client client = clients.getClient(fromUser, sourceOrganizationSid);
+        final String toUser = CallControlHelper.getUserSipId(request, useTo);
+        final Client toClient = clients.getClient(toUser, destinationOrganizationSid);
         final AccountsDao accounts = storage.getAccountsDao();
         final ApplicationsDao applications = storage.getApplicationsDao();
 
+        IncomingPhoneNumber number = getIncomingPhoneNumber(to, sourceOrganizationSid, destinationOrganizationSid);
+
+        if (number != null) {
```
jaimecasero: i notice this piece of code is heavily reuse/duplicated later. can we consider to place this in a common util...?

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/ua/UserAgentManager.java
```diff
@@ -419,6 +420,9 @@ private boolean permitted(final String authorization, final String method, SipUR
         final String response = map.get("response");
         final ClientsDao clients = storage.getClientsDao();
         final Client client = clients.getClient(user, OrganizationUtil.getOrganizationSidBySipURIHost(storage, sipURI));
+        if (client != null && storage.getAccountsDao().getAccount(client.getAccountSid()).getStatus() != Account.Status.ACTIVE) {
```
jaimecasero: in this case is mapped to 403/FORBIDDEN...

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/UssdPullTest.java
```diff
@@ -438,6 +440,40 @@ public void testUssdMessageLengthExceeds() {
...
+    public void testUssdPullClosedAccount() {
+        final SipCall bobCall = bobPhone.createSipCall();
+        bobCall.initiateOutgoingCall(bobContact, ussdPullDidClosedAccount, null, UssdPullTestMessages.ussdClientRequestBody, "application", "vnd.3gpp.ussd+xml", null, null);
+        assertLastOperationSuccess(bobCall);
+
+        assertTrue(bobCall.waitOutgoingCallResponse(5 * 1000));
+        int responseBob = bobCall.getLastReceivedResponse().getStatusCode();
+        if (responseBob == Response.TRYING) {
+            assertTrue(bobCall.waitOutgoingCallResponse(5 * 1000));
+            assertTrue(bobCall.getLastReceivedResponse().getStatusCode() == Response.NOT_ACCEPTABLE);
```
jaimecasero: lets try to use assertEquals instead of this. Is pretty bad when test is suddenly broken, and fail message doesnt provide a hint.

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
```diff
@@ -168,7 +160,13 @@ public void onReceive(final Object message) throws Exception {
             final String method = request.getMethod();
             if ("INVITE".equalsIgnoreCase(method)) {
                 check(request);
-                invite(request);
+                if (request.getContentType().equals("application/vnd.3gpp.ussd+xml")) {
```
jaimecasero: what is this new condition about? is this related to the original  story?
anyway, this seems a good place to use 415/Unsupported Media Type

restcomm/restcomm.sms/src/main/java/org/restcomm/connect/sms/SmsService.java
gvagenas: @jaimecasero No objection on the 403/Forbidden but I think we shouldn't provide reason phrase for security reasons, I think RC shouldn't provide hints that the account is not activated

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/UssdPullTest.java
gvagenas: Agree

restcomm/restcomm.ussd/src/main/java/org/restcomm/connect/ussd/telephony/UssdCallManager.java
gvagenas: @jaimecasero this is not new, the content type check was moved to there from another point of the same class.
Agree about the response code

PR #2819, opened 2018-02-13
SuspendAccountTree-Restcomm1220

jaimecasero:

Fixes #


restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/client/rcmlserver/RcmlserverApi.java
```diff
@@ -35,27 +33,43 @@
...
  *
  * @author otsakir@gmail.com - Orestis Tsakiridis
  */
 public class RcmlserverApi {
+
     static final Logger logger = Logger.getLogger(RcmlserverApi.class.getName());
 
     enum NotificationType {
-        accountClosed
+        accountClosed, accountSuspended,
```
gvagenas: @jaimecasero Apparently RVD needs to get notified for all state changes in the accounts, right?

jaimecasero: Well, i was checking the code before @otsakir  answered my challenge, and decided this may be applicable not only to CLOSED, but all states. Then orestis finally answered, and we decided to keep the notification only for CLOSED status. Anyway,i left the new implementation,as is using async http client, and we may potentially send notification about any status (more generic approach)

jaimecasero:
@gvagenas  thanks for reviewing. Ill take care of tests before merging.

jaimecasero:
thisi si latest testsuite results after fixing a issue with jenkinsfile, mostly inline with master.
https://cxs.restcomm.com/view/VVSSquad/view/ConnectCommunity/job/Restcom-CI-Multibranch/view/change-requests/job/PR-2819/5/testReport/

illl check some tests locally,and merge if appropiate

PR #2820, opened 2018-02-13
Restcomm-1739 Change profile document data type to work with both HSQLDB and MySQL

gvagenas:
This refer to #RESTCOMM-1739

Fixes #


jaimecasero:
so basically we have moved from byte[]/BLOB  (java/sql) to string/longvar. What is the limit for longvar in DB? This reminds we may need to check the content size to reject the request if is too large. WE may impl this easily with a reqFilter for all resources/endpoints anyway.

gvagenas:
@jaimecasero For HSQLDB LONGVARCHAR  max length will be 16M.
For MySQL LONGTEXT max length can be 4G so we should restrict that to 16M, I can change that to the SQL schema

PR #2828, opened 2018-02-15
Restcomm 1623 - Fix Dial Fork race condition

gvagenas:
**What this PR does / why we need it**: Dial Fork race condition when multiple clients sends 200 OK same time

Fixes # https://telestax.atlassian.net/browse/RESTCOMM-1623


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -1240,9 +1244,13 @@ private void onCallStateChanged(Object message, ActorRef sender) throws Transiti
                 //Do nothing
                 break;
             case RINGING:
-                if (is(forking)) {
-                    outboundCall = sender;
+                if (logger.isInfoEnabled()) {
+                    String msg = String.format("Got 180 Ringing from outbound call %s",sender);
+                    logger.info(msg);
                 }
+//                if (is(forking)) {
```
jaimecasero: i guess we dont need to track as outboundcall on ringing, this is to early to do that, we need to wait for the first 200OK coming, right.
why dont we remove this commented code?

gvagenas: right @jaimecasero we don't need to track outbound call on ringing. I will remove the commented code

PR #2860, opened 2018-03-07
removed unnecessary params from record #RESTCOMM-1758

maria-farooq:


maria-farooq:
test suit is green for this

PR #2862, opened 2018-03-08
Restcomm 1580 - Support Custom Header for Dial verb

gvagenas:
please review this as well https://bitbucket.org/telestax/telscale-restcomm/pull-requests/64/restcomm-1861/diff


restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsEndpointTest.java
```diff
@@ -191,6 +191,27 @@ public void testCreateAccount() {
...
 
+    @Test @Category(FeatureExpTests.class)
+    public void testCreateAccountWithInvalidChars() {
+
+        String username1 = "my?account@company.com";
+        String username2 = "my@account@company.com";
+        String username3 = "my=account@company.com";
+        String password = "RestCom1233@";
+
+        JsonObject createAccountResponse = RestcommAccountsTool.getInstance().createAccount(deploymentUrl.toString(),
```
jaimecasero: if we use "createAccountResponse" we may assert based on status response

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
```diff
@@ -802,6 +805,29 @@ private void validate(final MultivaluedMap<String, String> data) throws NullPoin
...
+        } catch (AddressException ex) {
+            String msg = String.format("Provided email address %s is not valid",emailAddress);
+            if (logger.isDebugEnabled()) {
+                logger.debug(msg);
+            }
+            throw new IllegalArgumentException(msg);
+        }
+
+        String clientLogin = data.getFirst("EmailAddress").split("@")[0];
+        for (char ch: ClientLoginConstrains.NOT_ALLOWED_CHARS) {
```
jaimecasero: can we move this code to ClientLoginCosntraints and reuse?=

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ClientsEndpoint.java
```diff
@@ -254,9 +255,15 @@ private void validate(final MultivaluedMap<String, String> data) throws RuntimeE
         } else if (!data.containsKey("Password")) {
             throw new NullPointerException("Password can not be null.");
         }
-        // https://github.com/RestComm/Restcomm-Connect/issues/1979
-        if (data.getFirst("Login").contains("@")) {
-            throw new IllegalArgumentException("Login contains invalid character: @ "+data.getFirst("Login"));
+        // https://github.com/RestComm/Restcomm-Connect/issues/1979 && https://telestax.atlassian.net/browse/RESTCOMM-1797
+        for (char ch: ClientLoginConstrains.NOT_ALLOWED_CHARS) {
```
jaimecasero: this should be reused from ClientLoginConstraints

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
```diff
@@ -386,16 +405,17 @@ protected Response putCall(final String accountSid, final MultivaluedMap<String,
         }
 
         CreateCall create = null;
+
         try {
             if (to.contains("@")) {
```
jaimecasero: why when To header contains "@", custom headers are not supported?

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -2419,10 +2438,10 @@ else if (CallManagerResponse.class.equals(klass) && !((CallManagerResponse)messa
                     }
                     if (call != null && callInfo != null) {
                         create = new CreateCall(e164(callerId(verb)), e164(callee), username, password, false, timeout(verb),
-                                CreateCallType.SIP, accountId, callInfo.sid(), statusCallback, statusCallbackMethod, statusCallbackEvent, mediaAttributes);
+                                CreateCallType.SIP, accountId, callInfo.sid(), statusCallback, statusCallbackMethod, statusCallbackEvent, mediaAttributes, null);
```
jaimecasero: i thouhgt customerHeaders was alredy supported  for SIP noun..., i expected to pass custome headers here as well... or is it taken from uri later? Seems different ways to pass custom headers in CreateCall msg

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
```diff
@@ -1981,11 +2016,19 @@ private void outboundToClient(final CreateCall request, final ActorRef sender) t
         }
     }
 
+    private SipURI addCustomHeadersForToUri (String customHeaders, SipURI to) {
```
jaimecasero: i think this is a bit tricky with many assumptions. I worry about runtimeExceptions like arrayindexoutofbounds, if passed string is not as expected, and then create leaks... Some additional checkings to make it safer and fail safe is appropiate would be better..

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/AccountsEndpoint.java
gvagenas: @jaimecasero Good suggestion, I will move it

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/ClientsEndpoint.java
gvagenas: Agree

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
gvagenas: @jaimecasero this is a call to a SIP URI and custom header are supposed to be already in the SIP URI if needed

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
gvagenas: @jaimecasero Right, for SIP URI already supports custom headers and seems too much to extract the custom headers from the SIP URI only to add them to the SIP URI later

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java
gvagenas: Actually, checking if null for `customHeader` has been done before calling the method. 

I will add an extra check for the custom header that should contain `=`

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
jaimecasero: why dont we pass customer headers for clients/number the same way, as uri params in to param, so header parsing is centralized in one place...

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/AccountsEndpointTest.java
gvagenas: @jaimecasero this is a JsonObject and doesn't hold the status response. In the given test case it will be null

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
gvagenas: @jaimecasero initially implementation was this way, but I reverted this because of https://telestax.atlassian.net/browse/RESTCOMM-1852.

If the `Client` or `Number` in the `CreateCall` request contains the custom headers, for example, `alice?X-custom-header1=1234` or `+151298989898?X-custom-header1=1234`, any extension that will be executed in the call flow will have to normalize the client/number and I strongly believe this is not responsibility of the extension.
Also, CallManager `outbound` method in case of Client, will have to normalize string to get the client.

For the above reasons, I decided its better to extract custom headers before passing the request to CallManager

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/util/ClientLoginConstrains.java
```diff
@@ -0,0 +1,33 @@
...
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>
+ */
+
+package org.restcomm.connect.commons.util;
+
+public class ClientLoginConstrains {
+    private static char[] NOT_ALLOWED_CHARS = {'?', '=', '@'};
+
+    public static boolean isValidClientLogin (final String login) {
```
jaimecasero: 👍

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -2419,10 +2431,10 @@ else if (CallManagerResponse.class.equals(klass) && !((CallManagerResponse)messa
                     }
                     if (call != null && callInfo != null) {
                         create = new CreateCall(e164(callerId(verb)), e164(callee), username, password, false, timeout(verb),
-                                CreateCallType.SIP, accountId, callInfo.sid(), statusCallback, statusCallbackMethod, statusCallbackEvent, mediaAttributes);
+                                CreateCallType.SIP, accountId, callInfo.sid(), statusCallback, statusCallbackMethod, statusCallbackEvent, mediaAttributes, customHeaders);
```
jaimecasero: 👍

PR #2880, opened 2018-03-19
Added inbound sms SDR event

agafox:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/SmsInterpreter.java
```diff
@@ -576,6 +576,7 @@ public void execute(final Object message) throws Exception {
             final SmsMessage record = builder.build();
             final SmsMessagesDao messages = storage.getSmsMessagesDao();
             messages.addSmsMessage(record);
+            getContext().system().eventStream().publish(record);
```
jaimecasero: i could not find any other occurence of eventStream usage. how is this supposed to be consumed? where is the subcriber regsitered?

agafox:
https://github.com/RestComm/Restcomm-Connect/blob/2d9d4d98a35ef4f5a489cabb93262ba4ba595122/restcomm/restcomm.sdr.api/src/main/java/org/restcomm/connect/sdr/api/SdrService.java#L38

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/SmsInterpreter.java
agafox: https://github.com/RestComm/Restcomm-Connect/blob/2d9d4d98a35ef4f5a489cabb93262ba4ba595122/restcomm/restcomm.sdr.api/src/main/java/org/restcomm/connect/sdr/api/SdrService.java#L38

agafox:
@jaimecasero are any comments left from your side?

maria-farooq:
@agafox do we need to add any test to cover this case?

maria-farooq:
i was thinking maybe we can cover that RC is publishing on desired events in RC side as well. but as Oleg said he will cover it on sdr service side. this pr is merged

PR #2893, opened 2018-03-30
Improved log statements and test case to reproduce issue

gvagenas:
Race condition couldn't be reproduced neither in testsuite not at staging with live test calls.
Patch contains improvements in logging statements for the call flow so we would better detect issues in the future.
Check comments at https://telestax.atlassian.net/browse/RESTCOMM-1932 for more details


jaimecasero:
this PR is ok to be merged

PR #2904, opened 2018-04-07
Restcomm 2497: implement suspend account

### LINKED ISSUE #2497: Implement new Account State `Suspended`
Currently we can only update an account to `Closed` state, which means all related resources (DIDs, Clients etc) will be removed and Account will be inactive.
From this state there is no way to get back to Active.

For this reason we need to implement the `Suspended` state which means, Account and account's resources (DIDs, Clients etc) cannot operate but are not deleted. At any time a `Suspended` account can move to either `Active` or `Closed` state

ddhuy:
Currently we can only update an account to `Closed` state, which means all related resources (DIDs, Clients etc) will be removed and Account will be inactive.
From this state there is no way to get back to `Active`.

For this reason we need to implement the `Suspended` state which means, Account and account's resources (DIDs, Clients etc) cannot operate but are not deleted. At any time a `Suspended` account can move to either `Active` or `Closed` state

Fixes #2497 

Following regression tests are added as @gsaslis recommend:
- No REST API request is allowed for a `SUSPENDED` account. Account cannot login, create calls, search DID etc.
- All resources (apps, clients, DIDs etc) are suspended and cannot be used. Clients cannot be registered, DID numbers return 403
- All sub-accounts of the account and their resources are also suspended and cannot be used.
- The Account, Sub-Accounts and all related resources still exist / are not removed from db.
- A suspended account can move to either `CLOSED` state or to `ACTIVE` state again.
- Moving to `ACTIVE` state means that Account and sub-accounts are active and can be used again, same for all related resources.
- Moving to `CLOSED` state means that Account, sub-accounts and related resources are closed forever and will be never used again. External resources like, purchased DID number will be released (so no more costs) but everything else (account, sub-account, resources) will remain in the DB for audit purposes.


ddhuy:
Hi @gsaslis ,
In current design, each time we update account status, we also update status of its sub-accounts
<script src="https://gist.github.com/ddhuy/44a4384259826b33b54761128fa0af9e.js"></script>


The account is suspended then activated again so the suspended/inactive/uninitialized sub-accounts will be activated too.
33177183-532e11ce-d094-11e7-8c34-7d8c963a4bd0

Do you have any idea for this situation?

gsaslis:
@ddhuy thanks for the new PR. This is much clearer. 

with regards to your particular question, I will say the same thing I had replied back then: https://github.com/RestComm/Restcomm-Connect/issues/2497#issuecomment-347610760

Plz do feel free to let me know if that's not clear or you need any further clarification there.

ddhuy:
@gsaslis yes. I understand your point. But with the new implementation in our source code. Seems that we always change the sub-account status whenever the parent account status is changed. Please have look at this function [switchAccountTree](https://gist.github.com/ddhuy/44a4384259826b33b54761128fa0af9e). It is called whenever we change one account status.

PR #2913, opened 2018-04-13
Ussd sdr support

DeerBeer:
Implementation of SDR support for USSD events


maria-farooq:
@agafox @DeerBeer did you guys compile this locally?

Issue #2895, opened 2018-04-02
RC ignore Action Dial verb or Next verb, If Callee send BYE or Callee is un-available or Busy


xhoaluu:
RC ignore Action Dial verb or Next verb, If Callee send BYE or Callee is un-available or Busy

PR #2916, opened 2018-04-16
#2895 Stop ringing tone before move to next verb when Dial verb fail …

xhoaluu:
In Dial Action verb, there is Play/Say verb. If callee in Dial Verb return back 480 Unavailable. RC send RQNT for stop ringing tone and RQNT for play verb really fast at the same time. When Media server return back NTFY for stoping ringing tone RQNT, RC understand this is NTFY for Playing verb that make RC move to next verb and finish Play/Say verb at this point

### DIFF
--- a/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
+++ b/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
@@ -1328,6 +1328,12 @@ public class VoiceInterpreter extends BaseVoiceInterpreter {
                     if (dialBranches != null && dialBranches.contains(sender)) {
                         dialBranches.remove(sender);
                     }
+                    if (dialBranches == null || dialBranches.size() == 0){
+                        // https://github.com/RestComm/Restcomm-Connect/issues/2895
+                        // Race condition If there is Play/Say Verb after dial verb, if the Dia verb is failing to call to the other,
+                        // NTFY to stop ringing tone also stop RQNT for playing/say verb. Better to stop it here.
+                        call.tell(new StopMediaGroup(), self());
+                    }
                     checkDialBranch(message,sender,action);
                 } else if (sender.equals(call)) {
                     fsm.transition(message, finished);


xhoaluu:
org.restcomm.connect.testsuite.telephony.DialRecordingTest.testRecordWithErrorOnRecordAction
Pass on local.

restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
```diff
@@ -1328,6 +1328,12 @@ private void onCallStateChanged(Object message, ActorRef sender) throws Transiti
                     if (dialBranches != null && dialBranches.contains(sender)) {
                         dialBranches.remove(sender);
                     }
+                    if (dialBranches == null || dialBranches.size() == 0){
+                        // https://github.com/RestComm/Restcomm-Connect/issues/2895
+                        // Race condition If there is Play/Say Verb after dial verb, if the Dia verb is failing to call to the other,
+                        // NTFY to stop ringing tone also stop RQNT for playing/say verb. Better to stop it here.
+                        call.tell(new StopMediaGroup(), self());
```
jaimecasero: is this goign to fix the raceCondition by proper state transition? or just send the stopTone(RQNT) ealier so the chances for race condition are lower?

xhoaluu: @jaimecasero send the stopTone(RQNT) ealier so the chances for race condition are lower.

jaimecasero: i thought so, we may need to introduce proper states to ensure consistent behavior. We need to discuss with rest of team.

xhoaluu: @jaimecasero The ticket has been created for longer than 1 month, Can we accept this fix. I will create jira ticket and The team should fix this later by correct state machine?

xhoaluu: Jira with logs: https://telestax.atlassian.net/browse/RESTCOMM-2087

xhoaluu: @jaimecasero The fix just working when enabled-200-ok-delay is set to true.
When enabled-200-ok-delay is set to false. The Fix is not working.

So I think We should fix this by corresponding state transition.

xhoaluu:
Closing this PR as there is another PR with prober fix.

PR #2918, opened 2018-04-18
Restcomm 1995

maria-farooq:


restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/SIPOrganizationUtil.java
```diff
@@ -50,6 +52,18 @@ public static Sid searchOrganizationBySIPRequest(OrganizationsDao orgDao, SipSer
             // try to get destinationOrganizationSid from toUri
             destinationOrganizationSid = getOrganizationSidBySipURIHost(orgDao, (SipURI) request.getTo().getURI());
         }
+        if (destinationOrganizationSid == null) {
+            // try to get destinationOrganizationSid from Refer-To
+            Address referAddress;
+            try {
+                referAddress = request.getAddressHeader("Refer-To");
```
jaimecasero: i was hpoing to do this in REFER logic so is actually only attempted there, but i guess the perf penalty is minimal if any.

i was just wondering whether refer-to should take preference if present to actual toHeader. I hope if they are both present,they will be probably the same. In my mindf or refer case, refer-to header shoudl be first option, but we dont have anything (spec,req) to enforce that...

this will surely fix the customer usecase anyway.

i guess is ok to merge if ci results are good

PR #2933, opened 2018-05-03
Support MD5 and ClearText for Client registration - RESTCOMM-2096

gvagenas:

Fixes #


restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/configuration/sets/MainConfigurationSet.java
```diff
@@ -63,4 +63,6 @@
     String getClientAlgorithm();
 
     String getClientQOP();
+
+    String getClearTextPasswordAlgorithm();
```
jaimecasero: is this name making sense?
i was expecting getDefaultPasswordAlgorithm() or soemthing like that...

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/util/DigestAuthentication.java
```diff
@@ -77,24 +77,24 @@ public static String response(final String algorithm, final String user, final S
...
-        validate(user, realm, password, nonce, method, uri, algorithm);
+        validate(user, realm, password, nonce, method, uri, authHeaderAlgorithm);
         String ha1;
 
-        if(!password2.isEmpty()){
-            ha1 = password2;
-        }else{
-            final String a1 = A1(algorithm, user, realm, password, nonce, cnonce);
-            ha1 = H(a1, algorithm);
+        if (clientPasswordAlgorithm.equalsIgnoreCase(RestcommConfiguration.getInstance().getMain().getClearTextPasswordAlgorithm())) {
```
jaimecasero: why do we need a configuration setting to set the token for ClearText algorithm?

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/configuration/sets/MainConfigurationSet.java
jaimecasero: do we really need to make the token to identify clearText password algortihm to be configurable?

restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/util/CallControlHelper.java
```diff
@@ -63,10 +64,15 @@ static boolean permitted(final String authorization, final String method, DaoMan
...
+        if (client != null && daoManager.getAccountsDao().getAccount(client.getAccountSid()).getStatus() != Account.Status.ACTIVE) {
+            return false;
+        }
+
         if (client != null && Client.ENABLED == client.getStatus()) {
-            final String password2 = client.getPassword();
-            final String result = DigestAuthentication.response(algorithm, user, realm, "", password2, nonce, nc, cnonce,
+            final String password = client.getPassword();
+            final String clientPasswordAlgorithm = client.getPasswordAlgorithm();
+            final String result = DigestAuthentication.response(authHeaderAlgorithm, user, realm, password,clientPasswordAlgorithm, nonce, nc, cnonce,
```
jaimecasero: will this work when "cleartext" is populated?

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/configuration/sets/MainConfigurationSet.java
gvagenas: @jaimecasero we have `String getClientAlgorithm();` which will return the default client password algorithm which in currently is `md5` and its configurable at `restcomm.xml`

I added the `String getClearTextPasswordAlgorithm();` in order to be able to get the `cleartext`  string from around the project using configuration instead of providing the string every time. In few words, to prevent tedious mistakes such as `cleartext`, `clear_text`, `clearText` etc.

restcomm/restcomm.commons/src/main/java/org/restcomm/connect/commons/util/DigestAuthentication.java
gvagenas: @jaimecasero see my previous comment, this is needed so we use the same string thoughout the project.

restcomm/restcomm.telephony.api/src/main/java/org/restcomm/connect/telephony/api/util/CallControlHelper.java
gvagenas: @jaimecasero yes, its already covered in the testsuite

abdulazizali77:
>Do we need to care about the rest of authentication in voice/sms/ussd side?
Jaime, no we dont need sms/ussd auth, this is only SIP Client authentication. so it only happens during SIP Register

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/ClientPasswordMigrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+        JsonObject clientsMigratedJsonObj = RestcommOrganizationsTool.getInstance().migrateClientsOfOrganization(deploymentUrl.toString(), superAdminAccountSid, superAdminAuthToken, org1);
+        assertNotNull(clientsMigratedJsonObj);
+
+        Gson gson = new Gson();
+
+        Map<String,String> migratedClients = new HashMap<>();
+        migratedClients = (Map<String,String>) gson.fromJson(clientsMigratedJsonObj, migratedClients.getClass());
+
+        assertNotNull(migratedClients);
+        assertEquals(4, migratedClients.size());
```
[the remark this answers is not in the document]
gvagenas: @jaimecasero The Clients endpoint will return a Client object that will have `password_algorithm=XXX`
Actually, this part is missing and I will change the converter now
The migrate method of the endpoint, will return the client id and the new algorithm as we discussed.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/ua/UserAgentManager.java
```diff
@@ -311,7 +306,7 @@ public void onReceive(final Object message) throws Exception {
                 } else if(authenticateUsers) { // https://github.com/Mobicents/RestComm/issues/29 Allow disabling of SIP authentication
                     final String authorization = request.getHeader("Proxy-Authorization");
                     if (authorization != null) {
-                      if (permitted(authorization, method, (SipURI) request.getFrom().getURI())) {
+                      if (CallControlHelper.permitted(authorization, method, storage, OrganizationUtil.getOrganizationSidBySipURIHost(storage, (SipURI) request.getFrom().getURI()))) {
```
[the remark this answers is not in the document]
gvagenas: @jaimecasero this is to remove the duplicated `permitted` method that we had at `UserAgentManager` and ` CallControlHelper`

gvagenas:
@jaimecasero @abdulazizali77 The CallControlHelper will take care of the authentication of UserAgentManager and the rest (CallManager, SmsManager etc).
At the `CallControlHelper` we use the `DigestAuthentication` to generate the proper response based on the client's password and password_algorithm.
The support for clear_text password is implemented at the `DigestAuthentication` which is common to `UserAgentManager`, `CallManager` etc. Testsuite covers this for all kind of traffic..

The fallback mechanism is implemented in the MybatisClientsDao, where if the `password_algorithm` column is missing, we assume it `cleartext` and as we discussed this is temporary patch that will be removed in the next iteration. And this is also covered in the testsuite

PR #2937, opened 2018-05-07
[WIP] Maven assembly of WAR + Docker image

### LINKED ISSUE #2474: What is the right way to build from source?
I was a little confused by finding both a `build.xml`, as well as a `pom.xml` in the project, so I was wondering whether I should be building using ant or maven...

I then came across the [building from source](http://documentation.telestax.com/connect/configuration/How%20to%20build%20Restcomm-Connect%20from%20source.html#build-from-source) guide, however that instructed me to create a new build script: 

> You will need to create a script to build Restcomm-Connect. Copy the script below into a bash script on your local directory at the root of the RestComm-Connect. In this tutorial, the script below is stored in the file restcomm-connect-build.sh.

I did create the script, and that did help me build the project, but then I realized there's already *another* bash script in the project root already (yeah, i know, i should have noticed!) , which is also titled `build.sh` - implying that's what I should be using. 

I am therefore a little confused... 

The `.travis.yml` file, does imply that the `build.sh` is the right script to use... But then again the `README.md` mentions `CloudBees`.. 

Clearly, something needs to be updated, so I'm wondering where I can help. 

In addition, I am wondering if the build process itself can be simplified. From what I have seen, this currently entails: 
* running a bash script, which: 
** invokes a maven plugin to set some versions, and then,
** invokes an ant build, which: 
*** invokes a maven build,
** and once that's finished, more maven plugins are used.

I am wondering if there is already an ongoing effort to just switch everything to maven and get rid of the ant build... ?

I could help. As long as there are no legacy / backward-compatibility reasons that are keeping this from happening, I'd be happy to look into this in more detail and open a PR.

gsaslis:
Largely, this PR includes cherry-picked work from the `restcomm1657` branch, where there is an ongoing effort to move from `ant` to `maven` for the generation of the WAR artifacts.

On top of that, this PR introduces 2 extra features: 

1. a new maven profile which builds a WAR file with JUST restcomm connect + management UI. (i.e. without mediaserver, RVD and WebRTC demo in the same WAR file). 
2. the docker image generation of both the "all-in-one" (`restcomm/restcomm`), as well as the "standalone" (`restcomm/connect`), ported over from the Restcomm-Docker repository. 
With this, that repository largely (if not entirely) becomes deprecated and reduces our tech debt.

Fixes #2474 

1. Please note that due to the cherry picks, this PR also includes the wildfly maven module. I am not sure if you would like that included. If not, I can exclude it, because I presume RC does not fully support wildfly yet (which is the work underway in `restcomm1657`). 
2. At the moment, the Restcomm Media Server has its own docker image and - afaik - so does RVD + WebRTC Demo. This is the reason for the 2 new features above, so we can run each application in its own container, rather than everything-in-one. 
3. The generated `restcomm/connect` docker image has *already been tested with the new mediaserver docker image*, and we can get a simple `+1234`  call running correctly. If there is a more extensive list of tests you would like to run, happy to help with that.


release/restcomm-assembly-as7/pom.xml
```diff
@@ -0,0 +1,257 @@
...
+                            </images>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <id>start</id>
+                                <phase>pre-integration-test</phase>
+                                <goals>
+                                    <!-- "build" should be used to create the images with the
+                                         artifact -->
+                                    <goal>build</goal>
```
jaimecasero: again i dont see the "build" goal being hooked to pre-integration-test. i think "package" would be better

release/restcomm-assembly-as7/pom.xml
```diff
@@ -0,0 +1,257 @@
...
+    <artifactId>assembly-as7</artifactId>
+    <packaging>pom</packaging>
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <sipservletapi.version>3.2.0-89</sipservletapi.version>
+        <docker.buildArg.CONNECT_VERSION>${project.version}</docker.buildArg.CONNECT_VERSION>
+    </properties>
+    <profiles>
+        <profile>
+            <id>all-in-one</id>
```
jaimecasero: what happens on maven:deploy when both profiles are enabled. Will both binaries will get deployed into nexus with different classifier?

release/restcomm-assembly-as7/pom.xml
```diff
@@ -0,0 +1,257 @@
...
+                                    <goal>unpack</goal>
+                                </goals>
+                                <configuration>
+                                    <artifactItems>
+                                        <artifactItem>
+                                            <groupId>org.mobicents.servlet.sip</groupId>
+                                            <artifactId>sip-servlets-as7</artifactId>
+                                            <version>${sipservletapi.version}</version>
+                                            <type>zip</type>
+                                            <classifier>assembly</classifier>
```
jaimecasero: classifier is the same as previous assembly, will this create collision if both profiles are enabled?

jaimecasero:
@gsaslis i dont think we will be able to merge this into master with all the wildfly10 stuff, specially all the changes in the pom files...

gsaslis:
@jaimecasero i agree. this is what i tried to highlight on my opening PR comment above. 
how would you propose we handle the wildfly10 stuff then? Should I just delete that module from this PR ?

i followed this approach primarily so i can cherry pick your commits and you can keep track of what's included.

jaimecasero:
@gsaslis wildfly10 new module is not really any issue. But along with the new module there were lots of changes to the existing pom files to remove tomcat depends and bring widlfly. Those changes in core project pom files to accomodate wildfly10 qare the ones i would try to exlucude from this PR...

release/restcomm-assembly-as7/pom.xml
gsaslis: @jaimecasero that is a fair point. 

btw, do you agree that it _should_ be bound to some phase, or would you prefer that we added the plugin and then explicitly invoke the `docker:build` maven target when we want to build a docker image?

release/restcomm-assembly-as7/pom.xml
gsaslis: @jaimecasero with regards to the docker images, that is not an issue, because we have `restcomm/restcomm` and `restcomm/connect` images. 

What would you propose for WAR binaries?

release/restcomm-assembly-as7/pom.xml
gsaslis: i get your point, but - to make sure I did get it right - this is not the right place to add that comment, right? 

(i mean this right here is referring to the sip-servlets dependency, not the artifact generation... )

gsaslis:
@jaimecasero ah, i see, so you're saying i should reverse commit this: https://github.com/RestComm/Restcomm-Connect/pull/2937/commits/a4df2e09a6fc5236ec4b53f111615f23de0dfdc3
?

release/restcomm-assembly-as7/pom.xml
jaimecasero: im looking into hook docker image generation into some phase, so its part of "regular" build, and invoked with regular maven goals

release/restcomm-assembly-as7/src/main/resources/descriptor_all-in-one.xml
```diff
@@ -0,0 +1,68 @@
+<assembly xmlns="http://maven.apache.org/ASSEMBLY/2.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.0.0 http://maven.apache.org/xsd/assembly-2.0.0.xsd">
+    <id>assembly</id>
```
jaimecasero: this is the "right" place i think. this id is same as standalone one, and could create potential collision

release/restcomm-assembly-as7/pom.xml
jaimecasero: right, this is not right place sorry, i guess you got the right point anyway..

jaimecasero:
@gsaslis  that commit a4df2e0 is specially bad as you seen, becuase it introduces wildfly dependencies, and modifies testsuite to run with wildfly.which is still far from supported...

release/restcomm-assembly-as7/pom.xml
gsaslis: 👍

release/restcomm-assembly-as7/pom.xml
jaimecasero: I guess you mean ZIP binaries... The assemble ZIP files should be deployed in nexus as attachment of project with specific classifier. But since you are using same project coordiante to create slightly different ZIP binaries, then assembly classifier is required to be different...

Honestly i tend to think these are really two different maven artifacts (all-in-one vs standalone),which should be modelled as separate maven modules with diff coordiantes(artifactid). In terms of zip versus docker, im fine sharing same project module, as their are semantically the same binary file (same features delivered) but with different flavors (ZIP versus Docker image). Also, we dont want to deploy docker image into Nexus as maven dependency,but be deployed in Nexus docker registry as docker image...

Again, im not sure about valid usage of profiles here...

release/restcomm-assembly-as7/src/main/resources/descriptor_all-in-one.xml
gsaslis: this should be addressed with https://github.com/RestComm/Restcomm-Connect/pull/2937/commits/71fd01f49b9dc91b099c0aaa1cbf9720fe5fb5ee

gsaslis:
@jaimecasero reverse committed with https://github.com/RestComm/Restcomm-Connect/pull/2937/commits/fe1e250f223686ce3d0d56f1fab3c524e489c4e7

PR #2950, opened 2018-05-15
Issue zendesk#35767, maxP2PCallLength configuration to extend P2P call, Ask partten for statuscallback

xhoaluu:

Fixes #

**What this PR does**
Point 1: Add configuration for Max P2P call length in restcomm.xml, that allow customer adjust when RC will release P2P within RC.
Point 2: Using ask partten for Completed statusCallBack of Dial verb. Some time, Call actor stop task too soon that make Downloader actor stop as well.

**why we need it**
https://telestax.atlassian.net/browse/RESTCOMM-2139
RC clear P2P call after 60 mins, that make Caller/Callee cannot send BYE to close the call with the other site.

https://telestax.zendesk.com/agent/tickets/35865
[Restcomm]Sometimes, StatusCallBack of "completed" is not created.

Customer confirm it fixed their issue.


restcomm/restcomm.application/src/main/webapp/WEB-INF/conf/restcomm.xml
```diff
@@ -266,6 +266,9 @@
 			 or send appropriate response according to the response from outbound dial-->
 		<!--  https://github.com/RestComm/Restcomm-Connect/issues/1649 -->
 		<enable-200-ok-delay>false</enable-200-ok-delay>
+        
+                <!-- Maximum P2P call length within RestComm, unit is minute -->
```
jaimecasero: do you expect this new configuration property to be configured for Docker container?. I think we are missing changes to autoconfig scripts there... not sure if you are planning to provide those later in appropiate repo

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Call.java
```diff
@@ -599,7 +604,18 @@ private void executeStatusCallback(final CallbackState state) {
 
                 if (parameters != null) {
                     requestCallback = new HttpRequestDescriptor(statusCallback, statusCallbackMethod, parameters);
-                    downloader.tell(requestCallback, null);
+                    if (!ask) {
```
jaimecasero: why do we make this optional? why dont we use the AskPattern always?

restcomm/restcomm.application/src/main/webapp/WEB-INF/conf/restcomm.xml
xhoaluu: @jaimecasero 
This configuration is for specific customer, we don't face with this requirement on cloud, so I'm not expecting to have automatic script do any job.

restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/Call.java
xhoaluu: @jaimecasero 
There are 4 stage: initial, ringing, answered and completed.

Just only completed state, Call actor will be close as soon as possible and don't wait downloader actor to finish the job. so we have to have some thing hold Call actor there until finish task.

Call actor is always alive for inital and ringing and answered state.

xhoaluu:
Rebase Branch with master. Rerun testsuite again, 4 faliling test are passed in local

org.restcomm.connect.testsuite.sms.push.SmsPushNotificationServerTest.testSmsEndpointMessage
org.restcomm.connect.testsuite.telephony.DialRecordingTest.testRecordWithErrorOnRecordAction
org.restcomm.connect.testsuite.UssdPullTest.testUssdPullClosedAccount
org.restcomm.connect.testsuite.telephony.DialRecordingTest.testDialClientAlice_AliceDisconnects

restcomm/restcomm.application/src/main/webapp/WEB-INF/conf/restcomm.xml
jaimecasero: i think is still required to add a new property here https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/advanced.conf

and then deliver this into the actual xml by  adding some scripting here 
https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/configuration/config-scripts/as7-config-scripts/restcomm/autoconfig.d/config-restcomm.sh#L718

jaimecasero:
test results look good,this is ready to merge

xhoaluu:
@jaimecasero I have added automation task for configure max-p2p-call-length. Please help me to review.

jaimecasero:
@xhoaluu  is this PR covering the fact container is not sending BYE to call participants when sessionTimer is expired?
assuming not..., should we cover this behavior on this PR, or this solution is enough for the customer?

jaimecasero:
@xhoaluu  as discussed i added commit to handle session expiration properly

gvagenas:
@jaimecasero as we discussed on the call, it might be necessary to check what is the application session timeout for akka actor `calls` because this patch might introducing an unexpected side-effect there. From what I see, the akka `call` actors are using the akka context receiveTimeout and not the sip timers.
I am also checking

gvagenas:
@jaimecasero Actually for some call flows we store the Akka `call` actor as an attribute in the application session, for example https://github.com/RestComm/Restcomm-Connect/blob/5292520c2667e8b6190d4c925222c5e5e0da8c5d/restcomm/restcomm.telephony/src/main/java/org/restcomm/connect/telephony/CallManager.java#L1403
but we don't set the sip timer as far as I can see, so there might be a side effect there

jaimecasero:
@xhoaluu  this has been merged, next releas will contain all fixes

PR #2956, opened 2018-05-25
jira-2165 RC should send BYE to XMS when stop recording.

xhoaluu:
RC does not send BYE to XMS if Record verb at the end, Caller drop the call at middle of record. That make XMS resource cannot be released.

Fixes #
https://telestax.atlassian.net/browse/RESTCOMM-2165
All information is provided in Jira ticket.

### DIFF
--- a/restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
+++ b/restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
@@ -67,6 +67,7 @@ import org.restcomm.connect.mscontrol.api.messages.StopMediaGroup;
 import org.restcomm.connect.mscontrol.api.messages.StopRecording;
 import org.restcomm.connect.mscontrol.api.messages.Unmute;
 import org.restcomm.connect.mscontrol.api.messages.UpdateMediaSession;
+import org.restcomm.connect.mscontrol.api.messages.RecordStoped;
 
 import javax.media.mscontrol.EventType;
 import javax.media.mscontrol.MediaEvent;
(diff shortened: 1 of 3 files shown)


restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
```diff
@@ -831,6 +832,7 @@ private void onStop(Stop message, ActorRef self, ActorRef sender) {
                 this.mediaGroup.getSignalDetector().stop();
                 this.collecting = Boolean.FALSE;
             }
+            call.tell(new RecordStoped(), self);
```
jaimecasero: do we need to provide smae change for non JSR309 controller?

jaimecasero:
@xhoaluu can you merge with latest master to check test results on CI again please? i think some of the failing tests were fixed on master

restcomm/restcomm.mscontrol.jsr309/src/main/java/org/restcomm/connect/mscontrol/jsr309/Jsr309CallController.java
xhoaluu: @jaimecasero 
Line 791:
https://github.com/RestComm/Restcomm-Connect/pull/2956/files/2772b7dc7d20cc069d1ae12ebb06e39f55019371#diff-57215364d1e805510727c150e410959bL791

This line of code hold the thread for more than 2 seconds every time RestComm want to stop recording from XMS.
When finish Record verb RC does:
 VoiceInterpreter send signal to stop recording to Jsr309CallController actor.
Jsr309CallController thread was block at line 791 for 2 second.
In the mean time Jsr309CallController is blocked, VoiceInterpreter send signal to close media session, but Jsr309CallController is blocked so the signal was ignored and then BYE cannot send to XMS.

In non JSR309, the code does not have any block point, so DELETE MGCP still send to RMS.

PR #2969, opened 2018-06-13
Support for Recording DELETE

gvagenas:

Fixes #

### DIFF
--- a/restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
+++ b/restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
@@ -24,7 +24,6 @@ import org.restcomm.connect.commons.amazonS3.S3AccessTool;
 import org.restcomm.connect.commons.configuration.RestcommConfiguration;
 import org.restcomm.connect.commons.dao.Sid;
 import org.restcomm.connect.core.service.api.RecordingService;
-import org.restcomm.connect.dao.DaoManager;
 import org.restcomm.connect.dao.RecordingsDao;
 import org.restcomm.connect.dao.entities.Recording;
 
@@ -35,17 +34,25 @@ public class RecordingsServiceImpl implements RecordingService {
 
     private static Logger logger = Logger.getLogger(RecordingsServiceImpl.class);
 
-    private final DaoManager daoManager;
+    private final RecordingsDao recordingsDao;
     private final S3AccessTool s3AccessTool;
+    private String recordingsPath;
 
-    public RecordingsServiceImpl (DaoManager daoManager) {
-        this.daoManager = daoManager;
-        s3AccessTool = daoManager.getRecordingsDao().getS3AccessTool();
+    public RecordingsServiceImpl (RecordingsDao recordingsDao, S3AccessTool s3AccessTool) {
+        this.recordingsDao = recordingsDao;
+        this.s3AccessTool = s3AccessTool;
+        this.recordingsPath = RestcommConfiguration.getInstance().getMain().getRecordingPath();
+    }
+
+    //Used for unit testing
+    public RecordingsServiceImpl (RecordingsDao recordingsDao, S3AccessTool s3AccessTool, String recordingsPath) {
+        this.recordingsDao = recordingsDao;
+        this.s3AccessTool = s3AccessTool;
+        this.recordingsPath = recordingsPath;
     }
 
     @Override
     public void removeRecording (Sid recordingSid) {
-        RecordingsDao recordingsDao = daoManager.getRecordingsDao();
         Recording recording = recordingsDao.getRecording(recordingSid);
 
         boolean isStoredAtS3 = recording.getS3Uri() != null;
@@ -55,8 +62,10 @@ public class RecordingsServiceImpl implements RecordingService {
                 s3AccessTool.removeS3Uri(recording.getS3Uri());
             }
         } else {
-            URI fileUri = URI.create(RestcommConfiguration.getInstance().getMain().getRecordingPath()+"/"+recordingSid+".wav");
-            File fileToDelete = new File(fileUri);
+            if (!recordingsPath.endsWith("/"))
+                recordingsPath += "/";
+            URI recordingsUri = URI.create(recordingsPath+recordingSid+".wav");
+            File fileToDelete = new File(recordingsUri);
 
             if (fileToDelete.exists()) {
                 fileToDelete.delete();
--- a/restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
+++ b/restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
@@ -19,9 +19,10 @@
 
 package org.restcomm.connect.core.service.recording;
 
+import akka.actor.ActorSystem;
 import org.apache.log4j.Logger;
+import org.joda.time.DateTime;
 import org.restcomm.connect.commons.amazonS3.S3AccessTool;
-import org.restcomm.connect.commons.dao.Sid;
 import org.restcomm.connect.core.service.api.RecordingService;
 import org.restcomm.connect.dao.DaoManager;
 import org.restcomm.connect.dao.RecordingsDao;
@@ -36,16 +37,17 @@ public class RecordingsServiceImpl implements RecordingService {
 
     private final DaoManager daoManager;
     private final S3AccessTool s3AccessTool;
+    private final ActorSystem actorSystem;
 
-    public RecordingsServiceImpl (DaoManager daoManager) {
+    public RecordingsServiceImpl (DaoManager daoManager, ActorSystem system) {
         this.daoManager = daoManager;
         s3AccessTool = daoManager.getRecordingsDao().getS3AccessTool();
+        this.actorSystem = system;
     }
 
     @Override
-    public void removeRecording (Sid recordingSid) {
+    public void removeRecording (Recording recording) {
         RecordingsDao recordingsDao = daoManager.getRecordingsDao();
-        Recording recording = recordingsDao.getRecording(recordingSid);
 
         boolean isStoredAtS3 = recording.getS3Uri() != null;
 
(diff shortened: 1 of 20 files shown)


restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+            URI fileUri = recording.getFileUri();
+            if (fileUri.getScheme().equalsIgnoreCase("file")) {
+                File recordingFile = new File(fileUri);
+                if (recordingFile.exists()) {
+                    recordingFile.delete();
+                }
+            }
+        }
+
+        recordingsDao.removeRecording(recordingSid);
```
jaimecasero: so its finally both physical (file removal) and logical (db removal),right?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/RecordingEndpointTest.java
```diff
@@ -128,6 +128,35 @@ public void getRecordingListFilteredByCallSid() {
         assertTrue(filteredRecordingsByCallSid.get("end").getAsInt() == 17);
     }
 
+    @Test
+    public void deleteRecording() {
+        JsonObject firstPage = RecordingEndpointTool.getInstance().getRecordingList(deploymentUrl.toString(), adminAccountSid,
```
jaimecasero: i think this first part about asserting the inital size of the list is not neccesary

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/RecordingEndpointTest.java
```diff
@@ -128,6 +128,35 @@ public void getRecordingListFilteredByCallSid() {
...
+        assertTrue(firstPage.get("end").getAsInt() == 34);
+        assertTrue(totalSize == 34);
+
+        //Recording SID to delete RE10000000000000000000000000000032
+
+        String recordingSid = "RE10000000000000000000000000000032";
+
+        RecordingEndpointTool.getInstance().deleteRecording(deploymentUrl.toString(), adminAccountSid, adminAuthToken, recordingSid);
+
+        firstPage = RecordingEndpointTool.getInstance().getRecordingList(deploymentUrl.toString(), adminAccountSid,
```
jaimecasero: i would just query for the recordign sid, and check that is 404,instead of basing the asserts on list

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import org.restcomm.connect.commons.dao.Sid;
+import org.restcomm.connect.core.service.api.RecordingService;
+import org.restcomm.connect.dao.DaoManager;
+import org.restcomm.connect.dao.RecordingsDao;
+import org.restcomm.connect.dao.entities.Recording;
+
+import java.io.File;
+import java.net.URI;
+
+public class RecordingsServiceImpl implements RecordingService {
```
jaimecasero: could we add unit testing for this service bean?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialRecordingS3UploadSecureTest.java
```diff
@@ -269,6 +274,116 @@ public synchronized void testDialClientAlice_BobDisconnects() throws Interrupted
 		verify(1, putRequestedFor(urlMatching("/s3/.*")));
 	}
 
+	@Test
+	@Category(FeatureAltTests.class)
+	public synchronized void testDialClientAlice_BobDisconnects_AndRemoveRecording() throws InterruptedException, ParseException, IOException {
```
jaimecasero: what is this test adding on top of the previous test in terms of purpose?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialRecordingTest.java
```diff
@@ -276,6 +278,93 @@ public synchronized void testDialClientAlice_BobDisconnects() throws Interrupted
         assertEquals(0, liveCallsArraySize);
     }
 
+    @Test
+    @Category(FeatureAltTests.class)
+    public synchronized void testDialClientAlice_BobDisconnectsAndRemoveRecording() throws InterruptedException, ParseException {
```
jaimecasero: i think we are adding too much testing for this feature, not sure what these tests are adding . I see maybe one of these tests mixing actual call with recording API usage, but 3 of them seems to much. I would prefer to invest more on unit testing,rather than integration testing

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialRecordingTest.java
```diff
@@ -601,6 +690,90 @@ public synchronized void testRecordVerb() throws InterruptedException, ParseExce
         assertEquals(0, liveCallsArraySize);
     }
 
+    @Test
+    @Category(FeatureAltTests.class)
+    public synchronized void testRecordVerbAndDeleteRecording() throws InterruptedException, ParseException {
```
jaimecasero: same comment as previous

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+        boolean isStoredAtS3 = recording.getS3Uri() != null;
+
+        if (isStoredAtS3) {
+            if (s3AccessTool != null) {
+                s3AccessTool.removeS3Uri(recording.getS3Uri());
+            }
+        } else {
+            URI fileUri = recording.getFileUri();
+            if (fileUri.getScheme().equalsIgnoreCase("file")) {
```
jaimecasero: what if scheme is not "file"? I would add some logging into this method

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+public class RecordingsServiceImpl implements RecordingService {
+
+    private static Logger logger = Logger.getLogger(RecordingsServiceImpl.class);
+
+    private final DaoManager daoManager;
+    private final S3AccessTool s3AccessTool;
+
+    public RecordingsServiceImpl (DaoManager daoManager) {
+        this.daoManager = daoManager;
+        s3AccessTool = daoManager.getRecordingsDao().getS3AccessTool();
```
jaimecasero: getting the tool from DAO seems a bit weird. was this already implemented this way? with the new service layer I think we should move this tool to service layer, and leave dao layer completely focus on persistance. Can we face this refactoring on this sprint?

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        Recording recording = recordingsDao.getRecording(recordingSid);
+
+        boolean isStoredAtS3 = recording.getS3Uri() != null;
+
+        if (isStoredAtS3) {
+            if (s3AccessTool != null) {
+                s3AccessTool.removeS3Uri(recording.getS3Uri());
+            }
+        } else {
+            URI fileUri = recording.getFileUri();
```
jaimecasero: this block of code handling the FS logic could be hidden behind an interface of maybe isolated in private method at least. If we choose interface, both S3 and FS mode should inherit same interface...

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
gvagenas: Right @jaimecasero

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/RecordingEndpointTest.java
gvagenas: @jaimecasero ok I can remove this

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialRecordingS3UploadSecureTest.java
gvagenas: @jaimecasero We need to test the following:
- Delete recording from Record verb
- Delete recording from Dial record
- Delete recording for recording uploaded to S3

And thus we have those 3 tests

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/telephony/DialRecordingTest.java
gvagenas: See my previous answer. I think 3 test cases is the minimum

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
gvagenas: @jaimecasero agree, I fixed that now

restcomm/restcomm.core/src/main/java/org/restcomm/connect/core/service/recording/RecordingsServiceImpl.java
gvagenas: Unit tests added

PR #2973, opened 2018-06-14
FS-56 Changed RVD context 'restcomm-rvd' -> 'visual-designer'

otsakir:
Replaced in source, tests, comments, SQL, config files and docs.
Skipped in old migration scripts.


jaimecasero:
@otsakir 
please merge with master to fix issues we identified in management module. at this moment this PR cant be compiled,so we cant check test results

Also, in general, assuming RVD is no longer part of RC docker image, why do we care what is the context-path in RC? Most of the changes i see are for autoConfig scripts, and testing resources. config ...

How is RC configured now in Cloud to use the independent RVD docker service?

otsakir:
@jaimecasero 

> @otsakir please merge with master to fix issues we identified in management module. at this moment this PR cant be compiled,so we cant check test results

When you say _management module_ do you refer to Console in community build ? If that's the case, i haven't enabled the change of context in the community branch.

> Also, in general, assuming RVD is no longer part of RC docker image, why do we care what is the context-path in RC? Most of the changes i see are for autoConfig scripts, and testing resources. config ...How is RC configured now in Cloud to use the independent RVD docker service?

Well, conceptually there are two changes. The change in the .war name and the change in the context path. We could assume that RC cloud wont' really care about the change in the .war name. So, the change in the .war name wouldn't make any difference. That may be the case. 

The second change is in the context path RC will search for RVD and this also affects cloud even though it uses RVD microservice.

We could get away with PRs with a smaller scope i believe. But then again, we would postpone potential problems and bugs for later when we forget about this. 

...

otsakir:
hmm, still need to update build.xml for community. I will do that now.

otsakir:
Created a local build and ran some manual tests. Everything seems to work fine.

PR #2977, opened 2018-06-18
Fix for #2976

### LINKED ISSUE #2976: Provide sorting in the backend for Calls Endpoint
Currently we only support client side sorting by date by using the Reverse=true/false parameter. The server only does paging. Let’s add support for sorting in the server side and also support for sorting by all currently displayed fields: From,To, Direction, Status, Duration, Price so the UX is generic. Also, remember to maintain default of sort date in descending order.

atsakiridis:
Improve UX for Console Logs/Calls screen

Fixes #2976 


restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/entities/CallDetailRecordFilter.java
```diff
@@ -89,6 +91,25 @@ public CallDetailRecordFilter(String accountSid, List<String> accountSidSet, Str
...
+        this.sortBy = mapAPIFieldsToDb(sortBy);
+        this.sortDirection = sortDirection;
+    }
+
+    // We want the fields that the API caller provides for sorting to have the exact same names as what
+    // is returned from the API when it returns CDRs in queries, so that the API is consistent and intuitive. However, when doing
+    // the actual query in the db we are using the DB column names which sometimes are different from those API-level
+    // fields. Let's use a mapping function to convert API-level fields to db column names.
+    public String mapAPIFieldsToDb(String sortBy) {
+        // TODO: One issue here is that whatever the API user sends inside SortBy is sent down to DB, which is messy. How is this currently handled in other cases in the API?
```
jaimecasero: i think any mapping to DB columns names should happen at DAO impl side, not in entities/RestEndpoint. So this is fair challenge

My suggestion is to add an specific sortByFrom sortByTo, and then check in DAO impl which ones are enabled, and set the query accordingly there.

jaimecasero: this is another bad example doing mapping in RESTEndpoint 
https://github.com/RestComm/Restcomm-Connect/blob/master/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/IncomingPhoneNumbersEndpoint.java#L341

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
```diff
@@ -82,11 +84,52 @@ public void getCallsList() {
         assertTrue(totalSize == 448);
     }
 
+    @Test
+    public void getCallsListUsingSorting() {
```
jaimecasero: is confirmed locally about test results https://cxs.restcomm.com/job/Restcom-CI-Multibranch/job/PR-2977/1/testReport/

could you please check those?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
```diff
@@ -82,11 +84,52 @@ public void getCallsList() {
...
+        // Provide both sort field and direction
+        // Provide ascending sorting and verify that the first row is indeed the earliest one
+        JsonObject response1 = RestcommCallsTool.getInstance().getCalls(deploymentUrl.toString(),
+                adminAccountSid, adminAuthToken, 0, 10, "date_created:asc", true);
+        assertTrue(((JsonObject)response1.get("calls").getAsJsonArray().get(0)).get("date_created").getAsString().equals("Fri, 5 Jul 2013 22:15:53 +0300"));
```
jaimecasero: i know this is well spread in the testsuite, but could you please switch this into proper assertEquals, then when assertion fails the message its much clearer. Please amend all occurences in this test class please.

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/entities/CallDetailRecordFilter.java
atsakiridis: @jaimecasero, I lost you, right now the mapping is done in the Filter; not in the entity/endpoint, or am I getting the organization wrong?

Also, if I'm getting this right, you are talking about introducing in the filter API level multiple sortBy* attributes, one for each column we want to support sorting on, and potentially make them enums with possible values ASC or DESC? And the parsing from the initial URL string to those attributes would still go the Endpoint code, right?

Then, when you mention DAO impl for the mapping, are you talking about MybatisCallDetailRecordsDao.getCallDetailRecords()? But there isn't the filter already initialized and ready for use by the mybatis mapper .xml file? I'm getting the feeling that it will be weird to update the filter there

An example from the API codebase where you use what you describe would also be great.

If it's easier we can also have a call if you want to further discuss

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: My bad, I was planning to check the results but totally forgot, will check them

atsakiridis: It's weird though because I ran the CallsEndpointTest.java multiple times with my changes and it was working fine. Anyway, let me get back to you on that

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: Sure, makes sense

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/entities/CallDetailRecordFilter.java
jaimecasero: my initial point is that i dont want to find DB column names in REST or Entity/DAOInterface level. The Filter you mention is part of the Entity layer in this case, so to me seeing DB column names there is bad.

Column names should be the responsability of DAO impl to me. If we can agree on that, I may elaborate on my suggested API changes...

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
jaimecasero: beware, soemtimes tests results depends on the current DB state, so running test cases alone, and runnig the full test class, may lead to different resutls.

Please ensure running the whole test class is providing stable results.

Also please notice the usage of  @FixMethodOrder(MethodSorters.NAME_ASCENDING) annotation in test class, to enforce a predictable order of test cases execution,and hence predictable results. Again reusing DB state from test case to test case is bad practice, but this is what it is now...

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: @jaimecasero fixed the issues in IT, and created PR to see if the CDR related ones are fixed. On the last failure on 'org.restcomm.connect.testsuite.sms.push.SmsPushNotificationServerTest.testSmsEndpointMessage' I think it's unrelated to my changes. Can you verify?

jaimecasero: right SmsPushNotificationServerTest is unstable test on master, we need to mark it unstable, but dont worry about this now... its not related to this PR

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
```diff
@@ -228,7 +236,82 @@ protected Response getCalls(final String accountSid, UriInfo info,
...
+                }
+                if (!sortDirection.equalsIgnoreCase("asc") && !sortDirection.equalsIgnoreCase("desc")) {
+                    return status(BAD_REQUEST).entity(buildErrorResponseBody("Error parsing the SortBy parameter: sort direction needs to be either \'asc\' or \'desc\'", responseType)).build();
+                }
+            }
+        }
+
+        ///////
+        if (sortBy != null) {
+            if (sortBy.equals("date_created")) {
```
jaimecasero: is "date_created" REST API parameter? i think is called "DateCreated" in REST API. the point of all this refactorign is to hide DB names at this layer...

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
```diff
@@ -228,7 +236,82 @@ protected Response getCalls(final String accountSid, UriInfo info,
...
+                if (sortDirection != null) {
+                    if (sortDirection.equalsIgnoreCase("asc")) {
+                        filterBuilder.sortedByDate(SortDirection.ASCENDING);
+                    } else {
+                        filterBuilder.sortedByDate(SortDirection.DESCENDING);
+                    }
+                }
+            }
+
+            if (sortBy.equals("from")) {
```
jaimecasero: is "from" or "From"?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
```diff
@@ -59,57 +61,100 @@ public void getCallsList() {
...
+        assertEquals(((JsonObject)response1.get("calls").getAsJsonArray().get(0)).get("date_created").getAsString(), "Fri, 5 Jul 2013 22:15:53 +0300");
+
+        // Provide only sort field; direction should default to desc
+        JsonObject response2 = RestcommCallsTool.getInstance().getCalls(deploymentUrl.toString(),
+                adminAccountSid, adminAuthToken, 0, 10, "date_created", true);
+        assertEquals(((JsonObject)response2.get("calls").getAsJsonArray().get(0)).get("date_created").getAsString(), "Tue, 31 May 2016 16:20:22 +0300");
+*/
+
+        JsonObject response3 = RestcommCallsTool.getInstance().getCalls(deploymentUrl.toString(),
+                adminAccountSid, adminAuthToken, 0, 10, "date_created:desc", true);
```
jaimecasero: is this the expected name "data_created"?
also ,lets try to check DAO behavior on unit tests first..

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
atsakiridis: @jaimecasero I agree on that, the problem is that the names I'm using are taken from what the API is returning for the CDR entity in the REST response not the DB:

```
    "calls": [
        {
            "sid": "CA011c9e9b9f80441a92eb6de49b0b9645",
            "InstanceId": "IDbb080f9ed7c34ac78197b35c12b58ab4",
            "date_created": "Fri, 16 Sep 2016 13:48:14 +0000",
            "date_updated": "Fri, 16 Sep 2016 13:48:14 +0000",
            "account_sid": "ACd239d5cab74caaa43a594772e4724687",
            "to": "jean",
            "from": "antonis",
            "status": "update_forced",
            "price_unit": "USD",
            "direction": "Client-To-Client",
            "api_version": "2012-04-24",
            "caller_name": "antonis",
            "uri": "/2012-04-24/Accounts/ACd239d5cab74caaa43a594772e4724687/Calls/CL2735f4b1998c4c6ab8944451503e53ba.json",
            "subresource_uris": {
                "notifications": "/2012-04-24/Accounts/ACd239d5cab74caaa43a594772e4724687/Calls/CA011c9e9b9f80441a92eb6de49b0b9645/Notifications.json",
                "recordings": "/2012-04-24/Accounts/ACd239d5cab74caaa43a594772e4724687/Calls/CA011c9e9b9f80441a92eb6de49b0b9645/Recordings.json"
            }
        },
```

And I think we should be using the same names in our API to make it more consistent and intuitive.

So I believe the root cause here is that some of the keys in the response are the same as the db columns, right?

atsakiridis: Yeah @jaimecasero , something is wrong I'm guessing in the converter because the xml response is fine:
```
        <Call>
            <Sid>CA01...</Sid>
            <InstanceId>IDbb...</InstanceId>
            <DateCreated>Fri, 16 Sep 2016 13:48:14 +0000</DateCreated>
            <DateUpdated>Fri, 16 Sep 2016 13:48:14 +0000</DateUpdated>
            <ParentCallSid/>
            <AccountSid>ACd23...</AccountSid>
            <To>jean</To>
            <From>antonis</From>
            <PhoneNumberSid/>
            <Status>update_forced</Status>
            <StartTime/>
            <EndTime/>
            <Duration/>
            <Price>0.00</Price>
            <PriceUnit>USD</PriceUnit>
            <Direction>Client-To-Client</Direction>
            <AnsweredBy/>
            <ApiVersion>2012-04-24</ApiVersion>
            <ForwardedFrom/>
            <CallerName>antonis</CallerName>
            <Uri>/2012-04-24/Accounts/ACd239.../Calls/CL2...</Uri>
            <SubresourceUris>
                <Notifications>/2012-04-24/Accounts/ACd239...7/Calls/CA011c9.../Notifications</Notifications>
                <Recordings>/2012-04-24/Accounts/ACd.../Calls/CA011.../Recordings</Recordings>
            </SubresourceUris>
            <Ring__duration/>
        </Call>
```

Let me do some digging

atsakiridis: @jaimecasero, yep it's the response converter, check [here](https://github.com/RestComm/Restcomm-Connect/blob/4cb3186007f31a469d6a4738af7f671f9937137d/restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/converter/AbstractConverter.java#L110), where is the json conversion, and right above is the xml which is fine. And this seems to be done generically and not just for CDRs. Any ideas why we are enforcing different property names? 

I was thinking that for the sorting fields we should use the documented attribute names (i.e. DateCreated, etc), which are the same with .xml, and I guess we can correct the .json return values when we come up with a new API version at some point. Wdyt?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
atsakiridis: @jaimecasero yes, please check my above answer on this.

Also, another thing that would be nice is to have the attribute names (like DateCreated) stored somewhere as constants so that we can reuse in multiple places (like above, in the converter, etc) and make sure we don't mess up the names. We could declare them for example in CallDetailRecord.java for example? Seems like the object model somehow. Wdyt?

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
atsakiridis: What I ended up doing here is introducing a bunch of constants in the CallsEndpoint, similar to what @jaimecasero did for the IncomingNumbers

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: Sure, will check unit tests too

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: Sure, will keep in mind

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: Cool

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/entities/CallDetailRecordFilter.java
atsakiridis: This is already fixed per our discussion the other day, so we 're good

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/common/SortDirection.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+package org.restcomm.connect.dao.common;
+
+/**
+ * Sorting direction for various API Queries
+ */
+public enum SortDirection {
+    NONE,
```
maria-farooq: not sure why we need SortDirection.NONE

maria-farooq: ok i see in following CDR filter class that none would mean not to sort by that field

maria-farooq: still i think it could be null so we can maybe remove NONE, wdyt?

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
```diff
@@ -59,57 +61,96 @@ public void getCallsList() {
...
+        assertEquals(firstPage.get("end").getAsInt(), 49);
 
         JsonObject secondPage = (JsonObject) RestcommCallsTool.getInstance().getCalls(deploymentUrl.toString(),
-                adminAccountSid, adminAuthToken, 2, null, true);
+                adminAccountSid, adminAuthToken, 2, null, null, true);
         JsonArray secondPageCallsArray = secondPage.get("calls").getAsJsonArray();
-        assertTrue(secondPageCallsArray.size() == 50);
-        assertTrue(secondPage.get("start").getAsInt() == 100);
-        assertTrue(secondPage.get("end").getAsInt() == 149);
+        assertEquals(secondPageCallsArray.size(), 50);
```
maria-farooq: thanks for fixing these assertions

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
```diff
@@ -49,22 +52,28 @@
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
+//import javax.ws.rs.WebApplicationException;
```
maria-farooq: lets remove if not required

maria-farooq:
apart from minor comments PR is ok from my side, you need green light on [ci job](https://cxs.restcomm.com/job/Restcom-CI-Multibranch/view/change-requests/job/PR-2977) and then its good to be merged

restcomm/restcomm.dao/src/main/java/org/restcomm/connect/dao/common/SortDirection.java
atsakiridis: Sure @maria-farooq, removing it

restcomm/restcomm.testsuite/src/test/java/org/restcomm/connect/testsuite/http/CallsEndpointTest.java
atsakiridis: Sure

restcomm/restcomm.http/src/main/java/org/restcomm/connect/http/CallsEndpoint.java
atsakiridis: Sure

PR #2978, opened 2018-06-19
BS-288 fixed inbound call is marked as canceled when creating bridge …

xhoaluu:
…in forking call

RCML:
<Response>
<Dial><Client>alice</Dial></Client>
<Dial><Client>bob</Dial></Client>
</Response>

1/ Call to the application above.
2/ bob answer the call.
3/ bob disconnect call
4/ Caller is not received BYE because bob already sent BYE.

Expectation: Caller should receive BYE from RC when bob sends BYE.

Investigation:
There is unknow race condition when RC receive onStateChanged(cancelled, outboudcall), RestComm use this state to mark inbound call state to be canceled. That make RestComm will not sent BYE to a canceled call when bob send BYE.

https://github.com/RestComm/Restcomm-Connect/blob/e6b299776d7e1532bfff5398e56b8ad494322488/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java#L1283

We should check if this is from LCM before marking this to canceled.

Fixes #
https://telestax.atlassian.net/browse/BS-288
This issue just happen on customer's environment.

### DIFF
--- a/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
+++ b/restcomm/restcomm.interpreter/src/main/java/org/restcomm/connect/interpreter/VoiceInterpreter.java
@@ -1278,7 +1278,7 @@ public class VoiceInterpreter extends BaseVoiceInterpreter {
                             removeDialBranch(message, sender);
                             checkDialBranch(message, sender, action);
                         }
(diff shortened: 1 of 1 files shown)


xhoaluu:
All failing testcases are passed in local machine.

PR #2985, opened 2018-06-29
BS-282 imcomming call duration should not contain Ringing duration of…

xhoaluu:
… outbound call

Incoming leg rings for 1m while waiting for the outbound leg to answer. We include that 1m on the incoming leg duration, but from Twilio example, it seems they don't do it.
And the biggest problem happens when outgoing leg doesn’t answer, in this case, basically, we charged the incoming leg with a ringing duration for a call that didn’t happen.

Fixes #
https://telestax.atlassian.net/browse/BS-282


gvagenas:
Merged

PR #2990, opened 2018-07-09
FS-298: Fixed wildcarding for body attribute

atsakiridis:
Currently, the Body filter doesn't apply properly and all rows are returned, no matter what it is. With this PR we make sure that it works and that partially matching search terms can return results properly, as is the case with From and to 

Fixes #

### DIFF
--- a/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/sms-messages.xml
+++ b/restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/sms-messages.xml
@@ -72,7 +72,7 @@
                     AND sender like #{sender}
             </if>
             <if test="body != null">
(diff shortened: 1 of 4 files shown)


restcomm/restcomm.application/src/main/webapp/WEB-INF/scripts/mariadb/sql/sms-messages.xml
```diff
@@ -72,7 +72,7 @@
                     AND sender like #{sender}
             </if>
             <if test="body != null">
-                    AND body like '%'|| #{body} ||'%'
+                    AND body like #{body}
```
maria-farooq: now that i am thinking
changing from `AND body like '%'|| #{body} ||'%'`
to `AND body like #{body}`
would make sql injection easier for a hacker
wdyt

can we still somehow concatenate `%` while working with sql

